PR #61, opened 2021-12-07
Update Polish translation

ChuckMichael:
The current translation is too long to fit in the button.
screenshot


MouriNaruto:
Thank you very much.

I wiil merge it after the NanaZip 1.0 stable is released because I have uploaded the package to the Microsoft Store and I haven't done the tag yet.

Kenji Mouri

PR #162, opened 2022-06-17
Allow NanaZip to be associated with any file type

manfromarce:
Starting from Windows 10 2004 (build 19041), if you add 
`<uap10:FileType>*</uap10:FileType>`
to the .appxmanifest file, the app can be selected in the "Open With" menu and in Properties>"Change..." for any file type.

This would allow users to associate NanaZip with less common archives which are actually ZIP files such as apk, cbz, jar, nupkg (I used to do the same with 7-Zip).


AndromedaMelody:
Thank you for your contribution.

I will merge this PR. Due to be compatible with Windows 10 (Build 17763~18363), I need to add `uap10` to `IgnorableNamespaces`.

PR #170, opened 2022-07-04
Enable security mitigations

dinhngtu:
7-Zip (and NanaZip by extension) often needs to parse multiple types of file, most of which originating from the Internet. There has been known vulnerabilities in the 7-Zip parser, some of them leading to remote code execution.

This pull request aims to enable several security mitigations in NanaZip against these vulnerabilities:
- Control Flow Guard and EH Continuation Metadata: provided by MSVC to mitigate ROP attacks;
- CET Shadow Stack: marks the NanaZip binaries as compatible with hardware shadow stack technologies in new CPUs; this feature was tested on a new AMD CPU with CET.
- Commit e4559fd specifically enables runtime process mitigation policies (as provided by Windows exploit protection):
  - Strict handle checks to block the use of invalid handles;
  - Dynamic code protection prevents the NanaZip process from generating code at run time (I manually verified that NanaZip does not do such things);
  - Image load policy blocking NanaZip from accidentally loading unexpected libraries from remote sources.

I couldn't find a comprehensive test suite for 7-Zip, but I did test the changes manually on some actions (creating zip files, extracting several compressed file types, etc.) and I did not encounter any crashes during my testing. If you have any such test suite, please tell me.

ARM systems only have runtime mitigations due to lack of MSVC support. Testing on ARM systems would be appreciated.

Here are some quick and dirty benchmark numbers before and after applying the patch:

Before

before

After

after


MouriNaruto:
Before I merge it, I think you should check these:

- I hope `Mitigations.h` and `Mitigations.c` should put to the `NanaZip.Shared` folder instead of `SevenZip` folder because I am adjusting the codebase in `SevenZip` folder.
- I think you also should to enable that for NanaZipSfxWindows and NanaZipSfxSetup projects.

Kenji Mouri

dinhngtu:
I've updated the code following your comments:
- `Mitigations.[ch]` have been moved to a separate project `NanaZip.Shared.Mitigations`.
- Runtime mitigations have been added for all 3 SFX projects, including `NanaZipSfxConsole`. An exception is `NanaZipInstaller`, which does not have EHCont enabled; otherwise it fails to build under Release x64 and I'm not quite sure why.
- `SetProcessMitigationPolicy` is now resolved at runtime.

MouriNaruto:
@dinhngtu 

> NanaZipInstaller does not have EHCont enabled for now; otherwise it fails to build under Release x64 and I'm not quite sure why.

It caused by VC-LTL msvcrt.dll mode, the customized VCRT I used. (NanaZip only use VC-LTL msvcrt.dll mode in SFX stubs and Installer because it needs to support Windows Vista and later.)

I will commit some changes to Mile.Cpp project in NanaZip project separately. So, I reverted all changes you made in `Mile.*` folders because it doesn't belong to NanaZip project. It's part of Mile.Cpp project, the common infrastructures used in projects created by myself.

Kenji Mouri

MouriNaruto:
Note: SfxConsole's entry point is in MainAr.cpp too. So, I removed all changes in SfxCon.cpp.

Kenji Mouri

Mile.Project.Windows/Mile.Project.Cpp.props
```diff
@@ -140,7 +140,6 @@
       <PreprocessorDefinitions>_DEBUG;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <Optimization>Disabled</Optimization>
       <RuntimeLibrary Condition="'$(MileProjectEnableVCLTLSupport)' == 'true'">MultiThreadedDebug</RuntimeLibrary>
-      <DebugInformationFormat>ProgramDatabase</DebugInformationFormat>
```
dinhngtu: @MouriNaruto 

This setting is needed in Debug mode because CFG is not compatible with the default DebugInformationFormat setting of '/ZI'.

MouriNaruto:
@dinhngtu 

It looks runtime mitigations is OK. I'm also testing commits in Mile.Cpp about the following content:

```
  <ItemDefinitionGroup Condition="'$(WindowsTargetPlatformMinVersion)' != '' and $([System.Version]::Parse('$(WindowsTargetPlatformMinVersion)').Build) >= 10240">
    <ClCompile>
      <ControlFlowGuard Condition="'$(Configuration)'=='Release'">Guard</ControlFlowGuard>
      <GuardEHContMetadata Condition="'$(Platform)'=='ARM64' Or '$(Platform)'=='x64'">true</GuardEHContMetadata>
    </ClCompile>
    <Link>
      <CETCompat Condition="'$(Platform)'=='Win32' Or '$(Platform)'=='x64'">true</CETCompat>
    </Link>
  </ItemDefinitionGroup>
```

Kenji Mouri

MouriNaruto:
@dinhngtu 

I will not enable `GuardEHContMetadata` for NanaZip because it's not compatible with VC-LTL msvcrt.dll mode. (NanaZip only uses VC-LTL msvcrt.dll mode in SFX stubs and Installer because it needs to support Windows Vista and later. But these reuse other static library projects.)

But `ControlFlowGuard` and `CETCompat` are OK for all projects, even with projects which need VC-LTL msvcrt.dll mode.

Kenji Mouri

MouriNaruto:
Thank you for your patch. I have merged it.

Kenji Mouri

MouriNaruto:
@dinhngtu 

Commits for enabling `ControlFlowGuard` and `CETCompat`

Mile.Cpp: https://github.com/ProjectMile/Mile.Cpp/commit/8260fa4013cc69b19e3bd8ab4788c8c6ffa69b33
NanaZip: https://github.com/M2Team/NanaZip/commit/a3790506b61a2ca57771a979a79131a7ea58b0d7

Kenji Mouri

AndromedaMelody:
I found a issue in debug version. I think we should disable security mitigations when developing if code defines `_DEBUG`.
Screenshot 2022-08-18 124830

MouriNaruto:
@AndromedaMelody 

I will solve it via "not enable dynamic code protection" when building the Debug target.

Kenji Mouri

MouriNaruto:
The `GuardEHContMetadata` is already enabled due to VC-LTL has been updated.

Also, NanaZip tried to enable `GuardSignedReturns`.

https://github.com/M2Team/NanaZip/commit/4f6610fd0ea353d77a1249cad5ad7c9b422e0b50

Kenji Mouri

MouriNaruto:
@dinhngtu 

All features in your PR are enabled. I hope you will be pleased with that.

Kenji Mouri

dinhngtu:
Thanks. I've been testing some extra mitigations like MSVC XFG. It might be worth adding those once they become stable in MSVC.

MouriNaruto:
@dinhngtu 

I have forgot to say that the GitHub source code repository of Mile.Project.Windows is https://github.com/ProjectMile/Mile.Cpp. You will see all changes about Mile.Project.Windows. I hope it will give some help to you.

Kenji Mouri

PR #171, opened 2022-07-04
Fix wrong default SFX stub file name.

dinhngtu:
The default SFX file name in UpdateGUI.cpp was not updated.


MouriNaruto:
Thank you for your patch. I have merged it.

Kenji Mouri

Issue #203, opened 2022-08-24
No "NanaZip" in the context menu when selecting files or packages, but shows up when selecting folders


TECHNIKVERBOT:
NanaZip 1.2
Windows 10 21H2 (Build 19044)

![image](https://user-images.githubusercontent.com/59102649/186355814-92231632-2787-4f68-986f-344c59b6de3a.png)
![image](https://user-images.githubusercontent.com/59102649/186355928-eb4f77bc-4e72-4504-8394-c71d9af31608.png)

MouriNaruto:
If you can't find NanaZip in the context menu, please restart all File Explorer processes via Task Manager. (Due to Microsoft can't promise all things defined in the AppX manifest are registered after the app is installed in store or command line. It looks like an asynchronous progress.)

Kenji Mouri

TECHNIKVERBOT:
Still nothing after restarting explorer.exe. :/

TECHNIKVERBOT:
Rebooted my machine, same deal...

MouriNaruto:
@TECHNIKVERBOT 

> And just for the heck of it, tried it with the middle option set, still nothing

That option is only for showing system context menu or not when you right click the file items in NanaZip's file manager itself.

Kenji Mouri

RayGicEFL:
The problem also exists in my computer and it seems that restarting explorer isn't the solution. It is elusive, sometimes the section appears and after a while it just gone.

FR0Z3:
This problem is also happening to me, I am using ExplorerPatcher running under Windows 11 Build 22000.856 and using Disable Windows 11 Context menu. NanaZip Version: 1.2.252.0
![Scrshot](https://user-images.githubusercontent.com/33184021/187217573-0cf75292-45ab-402d-8f4b-904d3d5084d0.png)

![image](https://user-images.githubusercontent.com/33184021/187216598-9e04238e-905b-41c8-a630-7fbe06dde72b.png)

dinhngtu:
@MouriNaruto I've started looking at this problem myself. I think the problem is in the NanaZipShellExtension code.

I replaced the entire NanaZipShellExtension.cpp with this file [PhotoStoreContextMenu](https://github.com/microsoft/AppModelSamples/blob/master/Samples/SparsePackages/PhotoStoreContextMenu/dllmain.cpp) and the context menu works on both files and directories:
[...]

Issue #193, opened 2022-08-01
No "NanaZip" in old style context menu on Windows 11, latest update.  


GarnetSunset:
![image](https://user-images.githubusercontent.com/13430434/182231388-158eedda-656d-4395-9fa1-895a400675ce.png)
Restarted explorer etc, no luck, any help? :O

fluentmoheshwar:
Windows Version? It's fine for me.

AndromedaMelody:
- Fix in #205 
- No new feedback for a long time, so we will close this issue. If the problem is found to appear again in Windows with the latest LCU, you can reopen(and comment) this issue or submit an issue.

Thanks for your support.

GarnetSunset:
Sorry I didn't reply! Slipped my mind!

PR #205, opened 2022-09-01
Workaround for NanaZip not appearing in classic context menu.

dinhngtu:
This is meant to be a fix for #193, #203 and other related issues.

Basically, when registering the shell extension via MSIX, Windows doesn't accept our current verb name "NanaZipShellExtension" for the "*" file type. The "Directory" file type and normal extensions (e.g. ".zip") work as normal. Registering our extension via [ExplorerCommandHandler](https://gist.github.com/dinhngtu/2be7222068241f6b0801e448dd8c442c) works normally with the same verb.

This patch changes the verb name of our shell extension to work around this issue.

I've tested this patch on the English version of Windows 10 21H2 and Windows 11. There might be some language-related behavior here so testing on other Windows languages would be appreciated.

### DIFF
--- a/NanaZipPackage/Package.appxmanifest
+++ b/NanaZipPackage/Package.appxmanifest
@@ -261,11 +261,13 @@
         </uap:Extension>
         <desktop4:Extension Category="windows.fileExplorerContextMenus">
           <desktop4:FileExplorerContextMenus>
+            <!-- Our verb name starts with "C" to work around a bug preventing our shell menu from appearing in the
+                 classic context menu. -->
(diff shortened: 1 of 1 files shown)


MouriNaruto:
What about using `CNanaZipShellExtension`? It's also started with the `C` letter.

Kenji Mouri

dinhngtu:
I tested it and it seems to work fine.

MouriNaruto:
Thanks you for your contribution. I have merged them.

Kenji Mouri

PR #206, opened 2022-09-01
Check 7z compression parameter validity upon start of compression.

### LINKED ISSUE #87: Unable to allocate enough memory
I was trying to compress the game files of Witcher 3 into a 7z archive, but after a few seconds NanaZip would show an error about being unable to allocate enough memory for this operation. 

- The uncompressed files are about 33GB
- i have 32GB of RAM
- I tried to do the same with the latest release of 7-zip, which did not have the issue. While compressing the files 7-zip only needed just above 1,8GB of RAM. 
- Tested using Windows store preview version on Windows 11.

dinhngtu:
Fixes #87. The code to automatically reload compression parameters from Registry didn't check for their validity, leading to invalid parameters being passed to the 7zip worker and therefore causing an out-of-memory error:

` a -i#7zMap[...] -t7z -mx=5 -md=4294967295b -ms=9223372036854775808b -mmt=4294967295 -sae -- [...]`

This patch validates the presence of each parameter before appending it to the parameter string.

### DIFF
--- a/SevenZip/CPP/7zip/UI/Common/CompressCall.cpp
+++ b/SevenZip/CPP/7zip/UI/Common/CompressCall.cpp
@@ -243,14 +243,15 @@ HRESULT CompressFiles(
         params += fo.Method;
       }
 
-      if (fo.Level)
+      /* Level = 0 is meaningful */
+      if (fo.Level != static_cast<UInt32>(-1))
       {
         params += " -mx=";
         ConvertUInt32ToString(fo.Level, temp);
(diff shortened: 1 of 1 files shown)


dinhngtu:
Done. Though I'm not sure about comparing with 0 since -1 is the only value used by 7zip to denote that a parameter is not present.

MouriNaruto:
Thanks you for your contribution. I have merged them.

Kenji Mouri

PR #209, opened 2022-09-03
MSI setup

dinhngtu:
This (draft) PR adds a MSI-based setup for NanaZip. A setup bundle is included to choose the correct architecture/version at install time.

A change was needed to disable `.pri` resources in the shell extension so that it would work outside of a MSIX package.

The current setup is at an early stage so feedback would be appreciated.

List of ideas to be done:

- [ ] Add support for ARM64 setups. (Requires WiX 3.14 which is not yet stable; I can try upgrading if this is absolutely required)
- [ ] Customize setup UI and branding.
- [ ] Add file type associations.
- [ ] Add NanaZip and its aliases to PATH.
- [ ] Find a way to create dual-purpose setup bundles (supporting both per-user and per-machine installations).

### DIFF
--- a/NanaZip.sln
+++ b/NanaZip.sln
@@ -128,6 +128,8 @@ Project("{930C7802-8A8C-48F9-8165-68863BCCD9DD}") = "NanaZipSetup", "NanaZipSetu
 		{9A119A76-97CB-4490-B8C2-651576CB9302} = {9A119A76-97CB-4490-B8C2-651576CB9302}
 	EndProjectSection
 EndProject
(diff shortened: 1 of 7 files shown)


MouriNaruto:
'.pri' resources will be mandatory in the future version of NanaZip. Because NanaZip will use XAML Islands to refresh the UI.

NanaZip will provide exe installer (You can see the empty reversed NanaZipInstaller project in the current stage.) after NanaZip is modernized by XAML Islands.

Kenji Mouri

PR #236, opened 2022-11-11
Opt out of dynamic code mitigation on main FM thread.

### LINKED ISSUE #231: Starting Nanazip doesn't open app/explorer window
Hi, I have some strange bug. 
I have Nanazip from Windows store, it added itself to the context menu and it works just fine using context menu options.
However, when I doubleclick archive or try to open Nanazip.exe... nothing happens. I got circle cursor for a moment and nothing happens. I don't have any 'explorer-like' window. 

I've reinstalled Nanazip, even cleaned a little of registry. Still nothing.

dinhngtu:
Fix #231.

ExplorerPatcher uses IAT hooking to intercept multiple shell API calls. This is not compatible with our dynamic code mitigation.

Enable dynamic code optout on the main FM thread to avoid startup crashes for now.

### DIFF
--- a/NanaZip.Shared/Mitigations.cpp
+++ b/NanaZip.Shared/Mitigations.cpp
@@ -50,6 +50,7 @@ EXTERN_C BOOL WINAPI NanaZipEnableMitigations()
     {
         PROCESS_MITIGATION_DYNAMIC_CODE_POLICY Policy = { 0 };
         Policy.ProhibitDynamicCode = 1;
+        Policy.AllowThreadOptOut = 1;
         if (!my_SetProcessMitigationPolicy(
             ProcessDynamicCodePolicy,
             &Policy,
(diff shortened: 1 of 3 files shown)


MouriNaruto:
I have known about that issue for a long time. But I take no action because I am worried about it can reduce the security.

I really hope we can find a better way. For example, refuse all dlls from untrusted origins.

Kenji Mouri

dinhngtu:
Hi,

This is indeed a weakening of NanaZip's security mitigations. However at the moment NanaZip crashes silently when encountering Explorer extensions with weird behavior, which is quite disconcerting to the user.

- There is [PROCESS_MITIGATION_BINARY_SIGNATURE_POLICY](https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-process_mitigation_binary_signature_policy), which should be able to block non-Store DLLs, but the behavior is quite annoying, as it causes an error message box to show every time a DLL is blocked. It is also quite indiscriminate in what it blocks (e.g. IMEs).
- There is Chromium's method of hooking NtMapViewOfSection, which is quite involved and probably overkill for what's needed.
- Failing that, IMO we should at least inform the users of the known incompatibility with ExplorerPatcher and other utilities that patch the system.

MouriNaruto:
OK, I think I should merge your PR.

But the usage of NanaZipThreadDynamicCodeAllow must be limited. (I will make some changes based on your work.)

Kenji Mouri

MouriNaruto:
It also fixes #235.

PR #245, opened 2022-11-25
Update russian translation

Blueberryy:


wallrik:
Interesting. I learned something new about the world today, by reading your PR. I learned that some languages use sideways double chevrons as quotation marks - something I should probably have been told in school! 😅 Good thing we have [Wikipedia](https://en.wikipedia.org/wiki/Guillemet) as adults!

Blueberryy:
Yeah i don't know why they didn't used these quotations in original translation but i added them.

MouriNaruto:
Thank you very much. I have merged them.

Kenji Mouri

PR #256, opened 2022-12-16
Use modern IFileDialog for folder picker dialog 

### LINKED ISSUE #251: Use new folder browser
The folder browser used by 7-Zip is still the pre-Vista style dialog created by `SHBrowseForFolder`:

Pre-vista style folder browser

[The code](https://github.com/M2Team/NanaZip/blob/b8dc18ff671c582a95ae8c4636a4d0923ac30869/SevenZip/CPP/Windows/Shell.cpp#L183) could be modernized to use the new folder browser (perhaps through [FolderPicker](https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers.folderpicker?view=winrt-22621)).

reflectronic:
Closes #251

Currently, I have implemented a shim for `SHBrowseForFolderW` that shows uses the Vista Common File Dialog. It does not understand every possible option that `SHBrowseForFolderW` takes, but it is good enough for the purposes of 7-Zip. It is not perfect, and there is one thing in particular that I am not too happy about:

7-Zip sets the dialog's default folder by setting in a custom message loop for the dialog, passing itself the path to the default folder as a string through `LPBROWSEINFOW.lParam`, and navigating to that path when the dialog is initialized. This method cannot work with IFileDialog, so this shim relies on the fact that 7-Zip always passes an `LPWSTR` to the message loop. If 7-Zip ever changes the way they use `SHBrowseForFolderW`, this shim will stop working correctly. If you have a better idea for how to implement this, I am all ears.

| Before | After |
|--------|-------|
| Pre-vista style folder browser     | New vista-style folder browser     |

### DIFF
--- a/NanaZip.Shared/ModernWin32FileBrowser.cpp
+++ b/NanaZip.Shared/ModernWin32FileBrowser.cpp
@@ -1,0 +1,93 @@
+﻿/*
+ * PROJECT:   NanaZip
+ * FILE:      ModernWin32FileBrowser.cpp
+ * PURPOSE:   Implementation for Modern Win32 File Browser
+ *
+ * LICENSE:   The MIT License
+ *
+ * DEVELOPER: reflectronic (john-tur@outlook.com)
+ *            Mouri_Naruto (Mouri_Naruto AT Outlook.com)
+ */
+
+#include <Windows.h>
+#include <ShlObj_core.h>
+#include <wrl.h>
+
+EXTERN_C PIDLIST_ABSOLUTE WINAPI ModernSHBrowseForFolderW(LPBROWSEINFOW uType)
+{
+    using namespace Microsoft::WRL;
+    ComPtr<IFileDialog> fileDialog;
+    if (FAILED(CoCreateInstance(CLSID_FileOpenDialog, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&fileDialog))))
+    {
+        return nullptr;
+    }
+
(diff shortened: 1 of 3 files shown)


MouriNaruto:
I think it's a clever idea to achieve the goal. It looks nice.

But if you can use C++/WinRT instead of WRL, I would be more satisfied.

Kenji Mouri

AndromedaMelody:
Thanks for your contribution. It looks great.

> But if you can use C++/WinRT instead of WRL, I would be more satisfied.

[WRL is now superseded by C++/WinRT.](https://learn.microsoft.com/en-us/cpp/cppcx/wrl/windows-runtime-cpp-template-library-wrl) NanaZip uses C++/WinRT Template Library to achieve the goal of modernization (e.g. [NanaZipShellExtension](https://github.com/M2Team/NanaZip/tree/main/NanaZipShellExtension)...). We recommend to use `winrt::com_ptr` (`#include <winrt/base.h>`) instead of `Microsoft::WRL::ComPtr`.

reflectronic:
I initially tried with C++/WinRT, but I ran into linker errors because the version from the NuGet package was conflicting with the version from the SDK. I realize that I just needed to edit the vcxproj and add `<MileProjectEnableCppWinRTSupport>true</MileProjectEnableCppWinRTSupport>` to fix this, so I will move it over to C++/WinRT now.

MouriNaruto:
Thank you for your contribution.

I have merged them.

Kenji Mouri

PR #368, opened 2024-01-17
Don't pass common compression props to handlers.

### LINKED ISSUE #342: The parameter is incorrect.
Using the latest NanaZip Preview and 'The parameter is incorrect' error message shows whenever I try to compress a folder with Lizard even though I left the parameters box empty. Same thing happens when trying to use fast LZMA2. Basically most of algorithms advertised. 

However when attempting the same thing in 7-Zip-ZS everything works as planned.

dinhngtu:
Fix #342.

7-Zip ZS sends all common compression parameters (`mt`, `memuse`) down to the individual handlers. However, this is broken in NanaZip (probably due to the 23.01 port?)

Don't pass anything except for the `mt` parameter to fix this issue.


MouriNaruto:
I agree that the issue is caused by porting to 7-Zip mainline 23.01. Igor changes a lot and no other major derivatives keep up the upstream at least that time.

Kenji Mouri

MouriNaruto:
I have merged them. Thank you very much.

Kenji Mouri

PR #380, opened 2024-02-08
Classic: Link with runtimeobject.lib.

dinhngtu:
NanaZip Classic and its shell extension still depend on WinRT `runtimeobject.lib`. Link them to this library.


MouriNaruto:
I need some work to remove the C++/WinRT dependency for NanaZip Classic.

But I will merge your PR. Thank you.

Kenji Mouri

PR #537, opened 2024-12-30
Always split path in CExtractDialog.

dinhngtu:
Even if the archive name checkbox in CExtractDialog is not checked, NanaZip appends the archive name to its extraction path anyway. Fix this issue by always splitting the extraction path in NanaZip.Windows.


MouriNaruto:
Before I merge this PR, I hope you will follow my two suggestions.

- Please save the source code files in UTF-8 with BOM to ensure that the Windows and MSVC toolchain read them in UTF-8.
- I hope we can add the NanaZip modification mark when modifying the 7-Zip mainline source code for better mainline synchronization maintenance which can reduce the regression. Here is an example:
https://github.com/M2Team/NanaZip/blob/27f148f4fe52fe419c486971fc213e84b4cc4a51/NanaZip.Core/SevenZip/CPP/7zip/UI/Console/Main.cpp#L128-L137

Kenji Mouri

PR #547, opened 2025-01-14
Smart Extraction: Bug fix

R-YaTian:
I forgot to add this code for Shell Extension...

### DIFF
--- a/NanaZip.UI.Modern/NanaZip.ShellExtension.cpp
+++ b/NanaZip.UI.Modern/NanaZip.ShellExtension.cpp
@@ -452,7 +452,7 @@ namespace NanaZip::ShellExtension
                 }
 
                 std::wstring Folder = BaseFolder;
-                if (this->m_CommandID != CommandID::ExtractHere)
+                if (this->m_CommandID != CommandID::ExtractHere && this->m_CommandID != CommandID::ExtractHereSmart)
                 {
                     Folder += SpecFolder;
                 }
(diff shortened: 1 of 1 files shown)


MouriNaruto:
Please keep the maximum 80 half-width chars per line for the non-7-Zip mainline part source code.

Kenji Mouri

R-YaTian:
> Please keep the maximum 80 half-width chars per line for the non-7-Zip mainline part source code.
> 
> Kenji Mouri

Done!

Remonli:
@R-YaTian  Smart Extraction works unexpected. 

If a compressed file contains only directories without any single files, when extracting,  it will extract these directories directly
into the current directory instead of creating a separate directory.

Could you help fix this ?

R-YaTian:
> @R-YaTian Smart Extraction works unexpected.
> 
> If a compressed file contains only a few directories without any single files, when extracting, it will extract these directories directly into the current directory instead of creating a separate directory.
> 
> Could you help fix this ?

I will look into it when my free time.

xd003:
> @R-YaTian Smart Extraction works unexpected.
> 
> If a compressed file contains only a few directories without any single files, when extracting, it will extract these directories directly into the current directory instead of creating a separate directory.
> 
> Could you help fix this ?

Can confirm this issue
If a archive contains 2 or more folders, smart extract should create a new folder with the same name as the archive and extract the folders inside it. Currently it extracts them directly in the working directory

Should i open a github issue for this ?

PR #554, opened 2025-01-26
Update Legacy.resw pt-BR Update the language to be more suitable

maisondasilva:
Update the language to be more suitable


maisondasilva:
@MouriNaruto

PR #579, opened 2025-02-18
Update Albanian translation

F1219R:


MouriNaruto:
I want to ask a maybe dumb question: Does SQ equals to Albanian language?

Kenji Mouri

MouriNaruto:
OK. I hope you can change the PR title which can use Albanian instead of SQ. It can help me to write the release notes when I release the new version.

Note: I will merge that after about three hours because I'm not at home which cannot touch the Nana series source code tree.

Kenji Mouri

PR #612, opened 2025-03-22
JANITORIAL: Fix a bunch of typos in comments

raziel-:
One typo in an error message


MouriNaruto:
For synchronizing from upstreams properly without regressions, you should not change the source code which inherited from upstreams like 7-Zip mainline and etc.

Kenji Mouri

raziel-:
half-noob here, where can i see (in the files or elsewhere) that they are pulled from upstream, please?

MouriNaruto:
> half-noob here, where can i see (in the files or elsewhere) that they are pulled from upstream, please?

All files you modified, except the ReadMe.md.

Kenji Mouri

raziel-:
ouch...so not many reasons to keep this pr alive

thanks for the heads up

PR #617, opened 2025-04-05
Add empty NanaZip.ExtensionPackage.

dinhngtu:
Add NanaZip.ExtensionPackage in preparation for #502.

### DIFF
--- a/NanaZip.ExtensionPackage/NanaZip.ExtensionPackage.Installer.iss
+++ b/NanaZip.ExtensionPackage/NanaZip.ExtensionPackage.Installer.iss
@@ -1,0 +1,35 @@
+﻿#define AppName "NanaZip Extension Package"
+#define AppPublisher "M2-Team"
+#define AppCopyright "© M2-Team and Contributors. All rights reserved."
+#define AppURL "https://github.com/M2Team/NanaZip"
+
+#ifndef AppVersion
+#define AppVersion "5.1.0.0"
+#endif
+
+[Setup]
+AppId={{42795434-AB1A-4197-A724-F13E08953DFC}
+AppName={#AppName}
+AppCopyright={#AppCopyright}
+AppVersion={#AppVersion}
+AppPublisher={#AppPublisher}
+AppPublisherURL={#AppURL}
+AppSupportURL={#AppURL}
+AppUpdatesURL={#AppURL}
+DefaultDirName={autopf}\{#AppName}
+DisableDirPage=yes
+DefaultGroupName={#AppName}
+DisableProgramGroupPage=yes
+VersionInfoVersion={#AppVersion}
+
+PrivilegesRequired=lowest
+PrivilegesRequiredOverridesAllowed=dialog
+OutputBaseFilename=NanaZip.ExtensionPackage_{#AppVersion}
+SolidCompression=yes
+WizardStyle=modern
+
+ArchitecturesAllowed=x64compatible or arm64
+ArchitecturesInstallIn64BitMode=x64compatible or arm64
+
+[Languages]
+Name: "english"; MessagesFile: "compiler:Default.isl"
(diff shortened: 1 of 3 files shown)


MouriNaruto:
Thanks. But I think it's necessary to add the check for ARM64 Windows because I don't want ARM64 Windows users to install the x64 version by accident. It will reduce lots of issues.

Kenji Mouri

dinhngtu:
Converted to Inno Setup for eventual universal extension package installer.

MouriNaruto:
Maybe we can use https://www.nuget.org/packages/Tools.InnoSetup instead of installing Inno Setup toolchain separately to make a better out-of-box build experience.

Kenji Mouri

MouriNaruto:
If you don't trust the unofficial one, you can also provide the InnoSetup toolchain in the NanaZip.ExtensionPackage\Tools folder, similar as https://github.com/M2Team/NanaRun/tree/main/Tools does.

Kenji Mouri

dinhngtu:
Done using Tools.InnoSetup 6.4.2.

NanaZip.ExtensionPackage/NanaZip.ExtensionPackage.Installer.iss
```diff
@@ -0,0 +1,32 @@
...
+#define AppPublisher "M2-Team"
+#define AppCopyright "Copyright (c) M2-Team and Contributors. All rights reserved."
+#define AppURL "https://github.com/M2Team/NanaZip"
+
+#ifndef AppVersion
+#define AppVersion "5.1.0.0"
+#endif
+
+[Setup]
+AppId={{42795434-AB1A-4197-A724-F13E08953DFC}
```
MouriNaruto, line 11: It seems the AppId's GUID format has some issues.

NanaZip.ExtensionPackage/NanaZip.ExtensionPackage.Installer.iss
```diff
@@ -0,0 +1,32 @@
...
+AppUpdatesURL={#AppURL}
+DefaultDirName={autopf}\{#AppName}
+DisableDirPage=yes
+DefaultGroupName={#AppName}
+DisableProgramGroupPage=yes
+VersionInfoVersion={#AppVersion}
+
+PrivilegesRequired=lowest
+PrivilegesRequiredOverridesAllowed=dialog
+OutputBaseFilename=NanaZip.ExtensionPackage-{#AppVersion}
```
MouriNaruto: Should use '_' instead of '-'.

NanaZip.ExtensionPackage/NanaZip.ExtensionPackage.Installer.iss
```diff
@@ -0,0 +1,32 @@
+﻿#define AppName "NanaZip Extension Package"
+#define AppPublisher "M2-Team"
+#define AppCopyright "Copyright (c) M2-Team and Contributors. All rights reserved."
```
MouriNaruto: Use "© M2-Team and Contributors. All rights reserved." instead.

MouriNaruto:
Also, there are some suggestions:

- Add the detect logic to ensure the 64-Bit only support.
- I hope the script will automatically update the build number.
- 
Kenji Mouri

NanaZip.ExtensionPackage/NanaZip.ExtensionPackage.Installer.iss
dinhngtu: This is due to a quirk of Inno Setup's bracket quoting rules (only the open bracket is quoted).

dinhngtu:
> Also, there are some suggestions:
> 
> * Add the detect logic to ensure the 64-Bit only support.

Done: I've set `ArchitecturesAllowed` and `ArchitecturesInstallIn64BitMode` in my new push.

> * I hope the script will automatically update the build number.

This is actually done in `NanaZip.ExtensionPackage/NanaZip.ExtensionPackage.Installer.proj`. The AppVersion here is purely a fallback.

PR #636, opened 2025-05-08
MOTW enforcement changes

### LINKED ISSUE #425: Propagate Zone.ID stream to all supported files by default
The default setting of Zone.ID propagation is set to _No_.
obraz

For better security, my proposition is to change default option to _Yes_. It will provide better protection against files downloaded from the Internet.

dinhngtu:
Fix #425.

- Set default MOTW mode (including NanaZip.Core) to Office files
- Add more unsafe file types (exe ps1 etc.) to MOTW extension list
- Add `HKLM\Software\NanaZip\Enforced` key for machine-wide enforcement of MOTW settings

- [x] Test

### DIFF
--- a/Documents/Policies.md
+++ b/Documents/Policies.md
@@ -1,0 +1,20 @@
+# NanaZip Policies
+
+NanaZip supports setting system-wide policies via creating Registry values in
+the key `HKLM\Software\NanaZip\Policies`.
+These policies override user settings.
+
+Following is the list of currently-supported policy options.
+
+## Supported policy values
+
+### Propagate Zone.Id stream
+
+This value controls Mark-of-the-Web (MOTW) propagation of archive files.
+
+- Name: `WriteZoneIdExtract`
+- Type: REG_DWORD
+- Value:
+    - `0`: No
+    - `1`: Yes (all files)
+    - `2`: Only for unsafe extensions (does not support all nested archives)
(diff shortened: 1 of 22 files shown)


MouriNaruto:
I hope the enforcement policy values can be put under the "HKLM\Software\NanaZip\Policies" key for a better user editing experience and the future policy additions, and I suggest the Mark-of-the-Web enforcement policy should be named as "MotwEnforcement" REG_DWORD value key, and the value should be the MOTW level (0, 1, 2).

Note: You know that I started to create a new dynamic module called NanaZip.ModernExperience, it's also the necessary step to implement a first run MOTW policy selection dialog. I think it's better to also provide a first run dialog to make people choice the suitable MOTW policy they hope.

Kenji Mouri

dinhngtu:
I can rename the key to Policies, but I don't think the value should be renamed, as it has the same meaning as the per-user option (at least for the case of `WriteZoneIdExtract`).

MouriNaruto:
Add some final notes for talking this topic on Discord:

Key: HKLM\Software\NanaZip\Policies
Value Key:WriteZoneIdExtract
Value Key's Value: keep the same as WriteZoneIdExtract

I think it's more better than my suggestions.

Kenji Mouri

dinhngtu:
- Turns out the kOffice setting does not handle nested archives. As a result, I've updated the defaults to kAll (all files).
- Renamed enforced setting key to Policies.
- Added policies docs and updated Readme (with a few features not mentioned before).

Waiting for testing next.

MouriNaruto:
I think you should make sure 80 characters per line for documents because some guys may want to read documents via terminals.


Kenji Mouri

PR #660, opened 2025-06-19
Use macros for dependency properties

dongle-the-gadget:


MouriNaruto:
Maybe you can reduce some macro usage via trying something similar in https://github.com/ProjectMile/Mile.Windows.Helpers/blob/main/Mile.Helpers/Mile.Helpers.CppWinRT.h and example in https://github.com/M2Team/NanaGet/blob/main/NanaGet/TaskItem.h and https://github.com/M2Team/NanaBox/blob/main/NanaBox/RdpClient.h.

Kenji Mouri

dongle-the-gadget:
@driver1998 mentioned this on Discord, but we deemed it impossible here due to the dependency property registration requirement. I did use that pattern for events.

Although that said I'll use the Event helper from Mile instead of copying from WIL.

MouriNaruto:
OK. I also hope you can learn something like https://github.com/M2Team/NanaZip/blob/main/NanaZip.Modern/AboutPage.h to simplify the definitions. Here are two screenshots to mention something I want:

image
image

Kenji Mouri

dongle-the-gadget:
One problem with that approach is that it could cause conflicts (this happened to CsWinRT once). What the file currently does is to add short namespace statements (i.e. `namespace wf = winrt::Windows::Foundation`). I think this way is more concise, but I'm open to suggestions.

MouriNaruto:
For C++, it may not be a big issue because the WinRT-related headers are for internal use. (People will use C++/WinRT to generate headers from WinMD files. I don’t think we should use NanaZip.Modern's WinRT header files outside the NanaZip.Modern.) So, the only distinct between WinRT and non-WinRT is enough, and it will be helpful to keep every line shorter than 80 half-width characters.

In my opinion, I don‘t think "wf" or something short will be better. I don't like abbreviations without useful meanings. (This is why I don't like POSIX API naming schemes.)

Kenji Mouri

MouriNaruto:
Maybe it's not a good idea for using the "using namespace XXX" in source files. Use "namespace XXX { }" like https://github.com/M2Team/NanaZip/blob/main/NanaZip.Modern/AboutPage.cpp may better.

Kenji Mouri

PR #666, opened 2025-06-30
Redesign progress dialog using XAML

dongle-the-gadget:


MouriNaruto:
We should make the generated resw readable.

Also, I think we should have a separate PR for the resource migration tool. Also, we need to have a feature which can remove the migrated resource items support for the legacy resw.

Kenji Mouri

dongle-the-gadget:
I think removing strings is probably too risky, as we might accidentally break other components that rely on it.

MouriNaruto:
> I think removing strings is probably too risky, as we might accidentally break other components that rely on it.

You right. But I think we should provide a separate tool (maybe called NanaZip.RemoveLegacyStringResources, you only need to provide a C# method to support remove legacy resource by specific resource id) for that. It's helpful for us to remove something in the future. (Although we don't need to remove legacy resource items at the current stage.)

Also, I hope the current NanaZip.ResourceMigrator project can be renamed to NanaZip.MigrateLegacyStringResources.

I hope you can send me two PRs for NanaZip.MigrateLegacyStringResources and NanaZip.RemoveLegacyStringResources first, because I hope the maintainer tools will be separate commits for better commit history.

I also hope you can have a PR like preparing the modern progress dialog string resources PR after the toolchain PRs but before processing this PR (of course, it need you think the string resources are ready). Just for better commit history.

Kenji Mouri

dongle-the-gadget:
> I also hope you can have a PR like preparing the modern progress dialog string resources PR after the toolchain PRs but before processing this PR (of course, it need you think the string resources are ready). Just for better commit history.

I think that we should combine it, as this is all under one functionality. Having the strings migrated before the functionality seems weird.

dongle-the-gadget:
I have created #668 to track the new legacy string migrator

MouriNaruto:
> > I also hope you can have a PR like preparing the modern progress dialog string resources PR after the toolchain PRs but before processing this PR (of course, it need you think the string resources are ready). Just for better commit history.
> 
> I think that we should combine it, as this is all under one functionality. Having the strings migrated before the functionality seems weird.

OK. You're right.

Kenji Mouri

PR #667, opened 2025-07-01
Improve Smart Extraction

### LINKED ISSUE #588: Issue with using Extract Here (Smart) with file names that exceed the path length limit
System information:
OS: Windows 11 Pro 24H2 26100.3194
Nanazip version 5.0.1263.0 x64

I'm having an issue when using the `Extract Here (Smart)` option on archives that contain files that have file names longer than Windows' path length limit.  With long path length file names `Extract Here (Smart)` behaves like the `Extract Here` option and does not create a folder to store the individual files in the archive.

For example, with this archive using `Extract to <Folder>` works as expected:

Image

However, `Extract Here (Smart)` does not create a single folder that stores the contents of the archive:

Image

Here are a couple of sample 7z archives that exhibit this behavior.
[samples.zip](https://github.com/user-attachments/files/18949641/samples.zip)

### LINKED ISSUE #591: Extract Here (Smart) doesn't work properly
System information:
OS: Windows 11 Pro 24H2 26100.2033
Nanazip version NanaZip 5.0 Update 2 (5.0.1263.0)  x64

Ideally, Smart Extract should work in such a way that if an archive contains two or more items (whether files or folders), it should create a new folder with the same name as the archive and extract the contents into it. However, I'm unable to replicate this behavior. When extracting an archive that contains two folders, the extraction happens directly in the current directory instead of creating a new folder and placing the extracted folders inside it. i am attaching a simple archive using which this scenario can be tested efficiently

[sample-1.zip](https://github.com/user-attachments/files/19041434/sample-1.zip)

R-YaTian:
Use std::set and wstring to solve issues.
This PR will fix #588 and close #591.

### DIFF
--- a/NanaZip.Core/SevenZip/CPP/7zip/UI/Common/Extract.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/UI/Common/Extract.cpp
@@ -16,6 +16,11 @@
 #include "Extract.h"
 #include "SetProperties.h"
 
+// **************** NanaZip Modification Start ****************
+#include <set>
+#include <string>
+// **************** NanaZip Modification End ****************
+
 using namespace NWindows;
 using namespace NFile;
 using namespace NDir;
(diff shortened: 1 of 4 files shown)


MouriNaruto:
~~All modifications in source code inherited from 7-Zip mainline must be add NanaZip Modification series comment marks.~~

It seems OK. I will merge that in recent days.

Kenji Mouri

PR #669, opened 2025-07-01
Add support for PyInstaller archive format

DevX-Cipher:
Implemented a native IInArchive handler to enable extraction of PyInstaller-packed executables. Integrated format detection and parsing logic into NanaZip core.


MouriNaruto:
Adding handlers which not exist in 7-Zip or 7-Zip ZS is not acceptable to add them to NanaZip.Core.

You should add your handler to NanaZip.Codecs without depending on 7-Zip mainline's definitions.

Kenji Mouri

DevX-Cipher:
@MouriNaruto please explain why

MouriNaruto:
The reason is obvious. NanaZip will only keep the Core Library and Self Extracting Executables implementation synchronized to the latest 7-Zip mainline implementation since NanaZip 3.0, all other things will be rewritten.

Kenji Mouri

MouriNaruto:
Note: NanaZip's specific handlers won't depend on 7-Zip mainline headers, but it's ABI compatible with 7-Zip mainline.

It's important for us to do the modernization work without busying on synchronizing the full mainline.

Kenji Mouri

DevX-Cipher:
Is this how your adding all other supports

MouriNaruto:
image
image

These are based on NanaZip.Codecs with simplified 7-Zip plugin interface definition header.

Kenji Mouri

DevX-Cipher:
@MouriNaruto i will check again today

DevX-Cipher:
@MouriNaruto this makes things more complicated then it already is

dinhngtu:
Following the new codec API is necessary so that new codecs are maintainable. The effort needed to port NanaZip for each 7-Zip release is already big as it is. The new codec is even ABI compatible with 7-Zip mainline so there shouldn't be any significant porting difficulties.

DevX-Cipher:
@dinhngtu are you able to implement pyinstaller archive

MouriNaruto:
Just read https://github.com/M2Team/NanaZip/blob/main/NanaZip.Codecs/NanaZip.Codecs.Archive.Ufs.cpp as one of examples of using 7-Zip plugin interface definitions sanitized by NanaZip.

Kenji Mouri

MouriNaruto:
The NanaZip's sanitized 7-Zip plugin interface definitions (which is only single C header file with Win32 COM style definitions) are able to use in both 7-Zip and NanaZip: https://github.com/M2Team/NanaZip/blob/main/NanaZip.Specification/NanaZip.Specification.SevenZip.h

In my opinion, I don't think using 7-Zip mainline's plugin interface definitions (which is messy, and nested with full of macros) is good for maintaining, even for the plugin developers, lol.

Kenji Mouri

DevX-Cipher:
@MouriNaruto if you decide to add it in let me know

DevX-Cipher:
@MouriNaruto any update

PR #687, opened 2025-07-21
Fix compression levels combo box

### LINKED ISSUE #673: No compression level 0 - store to turn off compression
NanaZip doesn't have the option to turn compress off. It's listed as **Level 1 (Store)** but I would of thought there would be a **Level 0** like 7-Zip has.
**NanaZip**

Image

**7-Zip**
Image
As you can see in the screenshot for 7-Zip below when compression **Level 0 store** is selected the **Dictionary size**, **Word size** and **Solid Block size** options are empty. NanaZip doesn't do this as though compression is still applied.

### LINKED ISSUE #685: Missing the "Copy" compression method in .7z format
Is this intentional or just a bug?

dongle-the-gadget:
Closes #673 
Closes #685

### DIFF
--- a/NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
+++ b/NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
@@ -305,6 +305,8 @@ struct CFormatInfo
 {
   LPCSTR Name;
 
+  // **************** NanaZip Modification Start ****************
+
   // List of levels supported by the format.
   // Format:
   // - Values are represented by left shifts of 1.
@@ -319,6 +321,9 @@ struct CFormatInfo
   // Typically, Store is Level 0.
   // Format is the same as the Levels field.
   UInt32 LevelsMask;
+
+  // **************** NanaZip Modification End ****************
+
   unsigned NumMethods;
   const EMethodID *MethodIDs;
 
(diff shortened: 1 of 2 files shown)


MouriNaruto:
Maybe we need to use the NanaZip modification mark to make us know that when we migrate this part to 24.09 in the near future.

Kenji Mouri

dongle-the-gadget:
The entire method is changed from mainline. Here's the original: https://github.com/ip7z/7zip/blob/395149956d696e6e3099d8b76d797437f94a6942/CPP/7zip/UI/GUI/CompressDialog.cpp#L1566-L1612

dinhngtu:
Could you give a summary?

dongle-the-gadget:
- `CFormatInfo.Levels` is to define which levels are supported by each format. Some formats have non-contiguous levels (i.e. ZIP has 0, 1, 3, 5, 7, 9)
- The `langID` change at the end of the diff hides the "Store" label on formats that don't support them. Convention is Store is always Level 0, except for Lizard and Zstd (hence the exclusion).

MouriNaruto:
I hope this summary can be noted as source code comments because maybe we will forget if only mention that in this PR's reply.

Kenji Mouri

dongle-the-gadget:
> I hope this summary can be noted as source code comments because maybe we will forget if only mention that in this PR's reply.

Done

MouriNaruto:
I forgot to say, the UI part is actually based on the 7-Zip 22.01. We need to notice that we ported the logic from the newer 7-Zip mainline in the comments with the following mark (It will be helpful to review the code after migrating UI to the newer 7-Zip implementation like 24.09.)

```
// **************** 7-Zip Backport Modification Start ****************
// **************** 7-Zip Backport Modification End ****************
```

I know there are many implementations without marks in the inherited 7-Zip mainline source code in UI components (except the SFX stubs). I know there will be lots of regression when I migrate. (But I hope to reduce the regression things.)

Kenji Mouri

dongle-the-gadget:
This is not really a backport, I made my own logic to deal with this.

MouriNaruto:
> This is not really a backport, I made my own logic to deal with this.

OK. It seems we should add these:

```
```

The modification mark is necessary to mention the modification. (This is why I suggest we should rewrite as more as possible UI related implementation instead of modifying on inherited implementation directly.)

For big modifications, you can follow the style like https://github.com/M2Team/NanaZip/blob/a912bc154f88aee4e392a9f72eac5559c5dc2c79/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHeader.h#L53-L86.

Kenji Mouri

dongle-the-gadget:
The modification marks have been added

PR #689, opened 2025-07-22
Initialize StoreContext later

dongle-the-gadget:
Improves startup time.

### DIFF
--- a/NanaZip.Modern/MainWindowToolBarPage.cpp
+++ b/NanaZip.Modern/MainWindowToolBarPage.cpp
@@ -157,7 +157,8 @@ namespace winrt::NanaZip::Modern::implementation
 
         std::wstring sponsorButtonLabel = L"[";
         sponsorButtonLabel += Mile::WinRT::GetLocalizedString(
-            L"NanaZip.Modern/MainWindowToolBarPage/SponsorButton/AcquireText");
+            L"NanaZip.Modern/MainWindowToolBarPage/SponsorButton/AcquireText",
+            L"Sponsor NanaZip");
         sponsorButtonLabel += L"]";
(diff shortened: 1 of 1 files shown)


NanaZip.Modern/MainWindowToolBarPage.cpp
```diff
@@ -155,13 +155,12 @@ namespace winrt::NanaZip::Modern::implementation
...
 
-        this->m_StoreContext = winrt::StoreContext::GetDefault();
-        if (this->m_StoreContext)
-        {
-            winrt::check_hresult(
-                this->m_StoreContext.as<IInitializeWithWindow>()->Initialize(
-                    this->m_WindowHandle));
-        }
+        std::wstring sponsorButtonLabel = L"[";
+        sponsorButtonLabel += Mile::WinRT::GetLocalizedString(
```
MouriNaruto, line 159: According to the definition of Mile::WinRT::GetLocalizedString.

https://github.com/ProjectMile/Mile.Windows.Helpers/blob/49dd11f06cd3e35e530ce32304dd687bf58c4c8d/Mile.Helpers/Mile.Helpers.CppWinRT.h#L103-L111

I think we should add the fallback string like "Sponsor NanaZip" which will show "[Sponsor NanaZip]". It will better than showing "[NanaZip.Modern/MainWindowToolBarPage/SponsorButton/AcquireText]" when failed to get the language resource.

Kenji Mouri

dongle-the-gadget: This is done in 074cd62

dinhngtu:
Can I have a Reported-by?

PR #702, opened 2025-08-05
Fix ASAR header padding

vapopescu:
ASAR format uses Pickle to serialize binary data. Pickle adds padding to align to INT32 if the size of the variable is not divisible by 4. We need to take this into consideration when reading the header string.


MouriNaruto:
It seems OK to me. I will merge that.

Update: I have tested the new logic modified by you and it's OK.

Kenji Mouri

PR #713, opened 2025-08-23
[Add] Auto extract all when opening .exe

MajThomas:
#712


dinhngtu:
Hello,

- Please modify both NanaZip.UI.Modern and NanaZip.UI.Classic.
- Please add the NanaZip modification markers to your changes like this: https://github.com/M2Team/NanaZip/blob/a4a620ff9b316dd66ded440cd509d9cd86153568/NanaZip.Core/SevenZip/C/Sha1.h#L1-L4

PR #717, opened 2025-08-24
Add Auto Extract All When Opening EXE

MajThomas:
#713  Repo on last PR. @dinhngtu


dinhngtu:
Why not force-push your branch to refresh the PR instead of opening a new PR each time?

PR #719, opened 2025-08-24
Add parse environment variables in addressbar

MajThomas:
Make it possible to parse environment variables such as `%appdata%`, `%temp%`, `%programfiles%`.
Before:
before
After:
after

### DIFF
--- a/NanaZip.Modern/AddressBar.cpp
+++ b/NanaZip.Modern/AddressBar.cpp
@@ -113,7 +113,17 @@ namespace winrt::NanaZip::Modern::implementation
 
         __super::OnApplyTemplate();
     }
-
+// **************** NanaZip Modification Start ****************
+// * Expands environment variables like %appdata%, %temp%, %programfiles% 
+    std::wstring ExpandEnvironmentVariables(const std::wstring& input)
+    {
+        wchar_t buffer[MAX_PATH * 4] = {};
+        DWORD ret = ExpandEnvironmentStringsW(input.c_str(), buffer, ARRAYSIZE(buffer));
+        if (ret == 0 || ret > ARRAYSIZE(buffer))
+            return input; 
+        return buffer;
+    }
+// **************** NanaZip Modification end ****************
     void AddressBar::OnTextBoxPreviewKeyDown(
         winrt::IInspectable const&,
         winrt::KeyRoutedEventArgs const& args)
@@ -135,7 +145,13 @@ namespace winrt::NanaZip::Modern::implementation
             }
             else
             {
-                auto submitArgs = winrt::make<AddressBarQuerySubmittedEventArgs>(Text(), nullptr);
+                // **************** NanaZip Modification Start ****************
+                // * Parse  EnvironmentVariables when hitting enter.
+                std::wstring rawPath = Text().c_str();
+                std::wstring expandedPath = ExpandEnvironmentVariables(rawPath);
+                auto submitArgs = winrt::make<AddressBarQuerySubmittedEventArgs>(
+                    winrt::hstring(expandedPath), nullptr);
+                // **************** NanaZip Modification End ****************
                 QuerySubmitted(*this, submitArgs);
             }
             args.Handled(true);


dinhngtu:
How does this work when navigating paths and archive names containing a %? Wouldn't this cause problems?

dongle-the-gadget:
My comments on this:
- `NanaZip.Modern\AddressBar` is, in my opinion, the wrong place for this change. I personally would do this at `CPanel::BindToPath`.
- You don't need to add the NanaZip modification marks at `NanaZip.Windows`, `NanaZip.Modern` and `NanaZip.Modern.FileManager`.

dinhngtu:
> * You don't need to add the NanaZip modification marks at `NanaZip.Windows`, `NanaZip.Modern` and `NanaZip.Modern.FileManager`.

@MajThomas I think it's good to explain that this is because the files that you modified are NanaZip-exclusive and do not belong to an upstream.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -72,10 +72,48 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
...
 }
+bool HasDriveLetter(const UString& path)
+{
+    return path.Len() >= 2 &&
+        ((path[0] >= L'A' && path[0] <= L'Z') ||
+         (path[0] >= L'a' && path[0] <= L'z')) &&
+        path[1] == L':';
+}
+
+UString ExpandFirstEnvInPath(const UString &path)
```
MouriNaruto: Environment instead of Env. Please never be lazy.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -72,10 +72,48 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
...
+        result += path.Ptr(secondPercent + 1);
+    return result;
+}
 
-HRESULT CPanel::BindToPath(const UString &fullPath, const UString &arcFormat, COpenResult &openRes)
+HRESULT CPanel::BindToPath(const UString &fullPath, 
+  const UString &arcFormat, COpenResult &openRes)
 {
-  UString path = fullPath;
+  UString path ;
```
MouriNaruto: Use "::" prefix for functions in the global namespace to avoid name conflicts with other namespaces if you are using C++.

There are similar issues.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -72,10 +72,48 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
 {
   return DoesNameContainWildcard(path.Ptr(NName::GetRootPrefixSize(path)));
 }
+bool HasDriveLetter(const UString& path)
```
MouriNaruto: Modification Mark is necessary.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -72,10 +72,51 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
...
+    if (firstPercent == std::wstring::npos)
+        return path;
+    int secondPercent = path.Find(L'%', firstPercent + 1);
+    if (secondPercent == std::wstring::npos)
+        return path;
+    UString envName;
+    for (int i = firstPercent; i <= secondPercent; ++i)
+        envName += path[i];
+    wchar_t buffer[MAX_PATH * 4] = {};
+    DWORD ret = ::ExpandEnvironmentStringsW(envName.Ptr(),
```
MouriNaruto: This definition code style is not OK. Must be this style:
```
static void ConvertSizeToByteUnitString(
    UInt64 Value,
    wchar_t* String)
```

NEVER BE LAZY!

There are several similar issues.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -72,10 +72,51 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
...
+     buffer, ARRAYSIZE(buffer));
+    UString expanded = (ret == 0 || ret > ARRAYSIZE(buffer))?
```
MouriNaruto: This code snippet layout style is not OK.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
MouriNaruto: Also have the extra spaces.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -68,14 +68,63 @@ void CPanel::SetToRootFolder()
...
+        return path;
+    int secondPercent = path.Find(L'%', firstPercent + 1);
+    if (secondPercent == std::wstring::npos)
+        return path;
+    UString environmentName;
+    for (int i = firstPercent; i <= secondPercent; ++i)
+        environmentName += path[i];
+    wchar_t buffer[MAX_PATH * 4] = {};
+    DWORD bufSize = ::ExpandEnvironmentStringsW(
+                   environmentName.Ptr(),
```
MouriNaruto: Extra indent for calling functions and methods is forbidden. Only add 4 spaces than before. Here is an example:
```
DWORD bufSize = ::ExpandEnvironmentStringsW(
        environmentName.Ptr(),
```

There are many similar issues in your PRs.

MajThomas: I can't see what is defined as ovelly indented by these examples. 4 spaces before what? Do you mean at the beginning of the line? Or do you mean align vertically with the return value? 
I know it's 4 spaces at the beginning of the line for function definitions. But for function calls with a return value, the standard is still ambigous to me.

MouriNaruto: No ambiguous because if you see the examples in that document. Also, we need to reduce extra spaces.

MouriNaruto: If you read Mile series library implementations, you will know what's wrong.

MouriNaruto: I think every fresh contributors (who want to try as lazy as possible) should read implementations in https://github.com/ProjectMile/Mile.Windows.Helpers/tree/main/Mile.Helpers line by line to know what is the suitable coding style for Nana and Mile series projects.

dongle-the-gadget: I think @MajThomas has a point though. The rule list and examples are not comprehensive. I think it's better if the rule is explicitly worded rather than implied through examples.

dongle-the-gadget: I think it's 4 spaces after the type declaration, so
```cpp
DWORD bufSize = ::ExpandEnvironmentStringsW(
    environmentName.Ptr(), // 4 spaces after DWORD
```

MouriNaruto: > I think @MajThomas has a point though. The rule list and examples are not comprehensive. I think it's better if the rule is explicitly worded rather than implied through examples.

I think maybe I should add the reference code examples in the rules document. (Because it's complex to express all rules by words, but as an reference example, it should be clearer than before.)

dongle-the-gadget: I can help with wording them if you want to.

MouriNaruto: > I can help with wording them if you want to.

You can send a PR to MouriDocs repo if you want to help. Also, huge thanks for that.

MouriNaruto: I have updated the coding style document https://github.com/MouriNaruto/MouriDocs/tree/main/docs/24 to provide a catch-all clause for C/C++ scenarios, which will reduce some issues and excuses. Also, it will be helpful to make a comprehensive word explainations.

MajThomas: @dongle-the-gadget @MouriNaruto
Quality control is not easy work. If we want to document all the quality standard officially, we'll probably need a team of experienced people in quality control field. I suggest we use feedback control instead of open loop control. Focus on documenting these problems we encountered today rather than trying to establish all the rules in advance.

dongle-the-gadget: The PR I opened is a direct response to the problems we encountered today. As I also added to the fix, the rule list is not going to be fully comprehensive.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -68,14 +68,79 @@ void CPanel::SetToRootFolder()
...
+    if (secondPercent + 1 <(int) path.Len())
+    {
+        expandedPath += path.Ptr(secondPercent + 1);
+    }
+    return expandedPath;
+}
+// **************** NanaZip Modification End ****************
+
+HRESULT CPanel::BindToPath(const UString &fullPath, 
+  const UString &arcFormat, COpenResult &openRes)
```
dongle-the-gadget: Some notes about the parameters here:
- Here there are multiple parameters, so each parameter should be on a separate line.
- The official order is `Type&` and `Type const&`, so I would recommend changing to that.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -68,14 +68,79 @@ void CPanel::SetToRootFolder()
 }
 
 
-static bool DoesNameContainWildcard_SkipRoot(const UString &path)
+static bool DoesNameContainWildcard_SkipRoot(
+            const UString &path)
```
dongle-the-gadget: Incorrect indentation? The second parameter should be 4 spaces from the leftmost position.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
MajThomas: This line is not meant to be modified. Maybe something wrong is happening during syncronizing files. I'll try to fix that.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
MajThomas: Same here. Unexpected modification.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
dongle-the-gadget: Your new methods should follow 2 though.

MajThomas: I do noticed the coding style note says about using const type as input, but I didn't even know const can be placed in reversed order like this. If not addressed, I wouldn't even notice that. Thank you for the tips.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -72,10 +72,73 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
 {
   return DoesNameContainWildcard(path.Ptr(NName::GetRootPrefixSize(path)));
 }
+// **************** NanaZip Modification Start **************
```
MouriNaruto: Add an empty line before NanaZip Modification Start in this scenario for better git diff result.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -72,10 +72,73 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
...
+        expandedPath += path.Left(firstPercent);
+    }
+    expandedPath += expandedFirstEvironment;
+    if (secondPercent + 1 <(int) path.Len())
+    {
+        expandedPath += path.Ptr(secondPercent + 1);
+    }
+    return expandedPath;
+}
+// **************** NanaZip Modification End ****************
```
MouriNaruto: Add an empty line after NanaZip Modification End in this scenario for better git diff result.

MouriNaruto:
According to https://github.com/MouriNaruto/MouriDocs/tree/main/docs/24, you will know:

> Use "PascalCase" for all identifiers unless really necessary.

Kenji Mouri

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -72,10 +72,73 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
...
 HRESULT CPanel::BindToPath(const UString &fullPath, const UString &arcFormat, COpenResult &openRes)
 {
-  UString path = fullPath;
+// **************** NanaZip Modification Start **************
+//UString path = fullPath;
```
MouriNaruto: This indent is not OK. The modification mark should also be indent in the block. It should be basic because everyone should know which is better when layouting the code lines.

MouriNaruto:
For my observation, it seems you wants to lazy on layouting your source code. You may think I can merge your changes with your "perfect" layouted source code like ⑨ a.k.a. Cirno style. But I think you are dreaming like my junior colleagues, lol.

Also, every thing I noted in the document and reference style source code is important. I don't want to say lots of things repeatly.

But, there is a good news, this commit's coding style maybe I can see it will be OK to merge if you finish several changes in recent days.

Kenji Mouri

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
MajThomas: I think this is a matter of perspective. Identing the modification markers to align with the modified code sure looks better when you're paying attention to the modified code. But on the other hand, it is a modification marker, therefore aligning it with the original code shows clearer where the modification is inserted. Without receiving this specific instruction, I wouldn't take for granted which perspective is better.

MouriNaruto: Everyone should know respect with the "code block" structure is important to readability.

MajThomas: Yes, now I know the modification marker should be a part of the code block. I'll respect this rule and fix the problems when I have the time.
Also, if there's any further suggestions, please provide. I'll try to fix them all at once. And in the next PR perhaps, we can avoid these problems.

MouriNaruto: Maybe you should fix all of my comments told first.

NanaZip.UI.Modern/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -81,9 +81,69 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
   return DoesNameContainWildcard(path.Ptr(NName::GetRootPrefixSize(path)));
 }
 
+bool HasDriveLetter(UString const& path)
```
MouriNaruto: It seems you forgot to do changes for this file.

MouriNaruto:
Maybe this PR implementation can also be simplified.

According to the document of ExpandEnvironmentStringsW, you will know:

> "lpSrc" parameter: **A buffer that contains one or more environment-variable strings** in the form: %variableName%. For each such reference, the %variableName% portion is replaced with the current value of that environment variable.

Kenji Mouri

MouriNaruto:
It seems you also forgot to add the modification mark in the modern part's inherited 7-Zip source code.

Kenji Mouri

MajThomas:
> Maybe this PR implementation can also be simplified.
> 
> According to the document of ExpandEnvironmentStringsW, you will know:
> 
> > "lpSrc" parameter: **A buffer that contains one or more environment-variable strings** in the form: %variableName%. For each such reference, the %variableName% portion is replaced with the current value of that environment variable.
> 
> Kenji Mouri

You're right. Using drive letter to decide whether to expand seemed too complicated, and if the path string is a relative path, it doesn't solve the <%variableName% in file name> problem. Expand the first environment variable when the path string starts with a '%' should be a simpler way.

MouriNaruto:
@MajThomas 

I think maybe some guys want to use something like `C:\Users\%USERNAME%`.

For NanaZip's address bar scenario, it should not contain the relative path.

Kenji Mouri

dongle-the-gadget:
The scenario you gave is not very practical because there already exists `%USERPROFILE%` which points directly to the user directory.

MouriNaruto:
> The scenario you gave is not very practical because there already exists `%USERPROFILE%` which points directly to the user directory.

This is only an example, but the scenario type I mentioned should have many usage chances, and Windows File Explorer also supports that.

Kenji Mouri

MajThomas:
> @MajThomas
> 
> I think maybe some guys want to use something like `C:\Users\%USERNAME%`.
> 
> For NanaZip's address bar scenario, it should not contain the relative path.
> 
> Kenji Mouri

That's interesting. Can one create a file named %username% in c:/users? I don't have a computer by me, if so, how would explorer.exe handle this situation?

MouriNaruto:
@MajThomas 

It seems Windows File Explorer addressbar will ignore '%username%' file, which will go to your user folder, lol.

Kenji Mouri

MouriNaruto:
It seems the coding style is OK. Only left an implementation issue we talked about before.

Kenji Mouri

MajThomas:
> It seems the coding style is OK. Only left an implementation issue we talked about before.
> 
> Kenji Mouri

I've done some searching, it seems like  microsoft did'nt notice this %xxx% in filename problem 40 years ago. Maybe we shouldn't try to fix that from the first place, making things work different from cmd.exe and explorer.exe.

MouriNaruto:
@MajThomas

It's meaningless not to support the scenario I mentioned because some users want to use their custom environment variables to do similar things. Respecting Windows current behavior, for NanaZip's address bar, is OK.

Kenji Mouri

MouriNaruto:
Also, even just not respecting Windows behavior, the `ExpandFirstEnvironmentInPath` should also be simplified with `ExpandEnvironmentStringsW` API behavior mentioned in the Microsoft document.

Kenji Mouri

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/PanelFolderChange.cpp
```diff
@@ -75,7 +75,26 @@ static bool DoesNameContainWildcard_SkipRoot(const UString &path)
 
 HRESULT CPanel::BindToPath(const UString &fullPath, const UString &arcFormat, COpenResult &openRes)
 {
-  UString path = fullPath;
+  
+  // **************** NanaZip Modification Start **************
+  //UString path = fullPath;
+  wchar_t Buffer[MAX_PATH * 4] = {};
```
MouriNaruto: What's meaning of MAX_PATH * 4? I think 32767 will be better because it's the NT and Win32 UNC path's maximum length limit.

I think you can get some improvement hints from the following implementation, which is referenced in https://github.com/M2Team/NanaBox/blob/main/NanaBox/Utils.cpp

```cpp
std::wstring GetCurrentProcessModulePath()
{
    // 32767 is the maximum path length without the terminating null character.
    std::wstring Path(32767, L'\0');
    Path.resize(::GetModuleFileNameW(
        nullptr,
        &Path[0],
        static_cast<DWORD>(Path.size())));
    return Path;
}
```

dongle-the-gadget: You'd need to use 32768 due to null-terminator.

MouriNaruto: > You'd need to use 32768 due to null-terminator.

For char and wchar_t scenario, it need 32768. For the std::string and std::wstring scenario, it actually only need 32767 because null terminator will be included in that, which will actually alloc 32768 chars' memory space.

MouriNaruto:
In the current stage, this PR basically looks good to me.

@dinhngtu @dongle-the-gadget I want to know your opinions before I merge that this evening in my timezone.

Kenji Mouri

MouriNaruto:
@MajThomas

I forgot to say that I think it's necessary to use std::wstring because it can reduce the stack overflow issues. (Of course, resize the std::wstring size after calling the `ExpandEnvironmentStringsW ` API.)

Kenji Mouri

dongle-the-gadget:
I don't believe resizing to be a good choice since that adds additional copying overhead, and given that the only requirement for UString is a null-terminated string, I don't see this overhead being worth it.

If you wish to not allocate that many characters in the stack I suppose you could malloc (heap allocate) it instead.

MouriNaruto:
> I don't believe resizing to be a good choice since that adds additional copying overhead, and given that the only requirement for UString is a null-terminated string, I don't see this overhead being worth it.

I think the overhead is acceptable for making reusable implementations. I believe some guys may want to copy the code snippet to use directly from this project.

We also need to reduce the chance of stack overflow caused by big stack buffer. Maybe some one need to calling it nested in multiple levels like enumerating folders.

Kenji Mouri

MouriNaruto:
> If you wish to not allocate that many characters in the stack I suppose you could malloc (heap allocate) it instead.

I think std::string will have better readability for that goal. Use `malloc` and `strcpy_s/stdcat_s` looks not simple. (Maybe sometimes someone will forgot to `free` which can cause memory leak.)

Kenji Mouri

dongle-the-gadget:
From what I read from the Internet, std::string should be heap allocated unless if the string is known to be small (which it isn't for this case)

MouriNaruto:
> From what I read from the Internet, std::string should be heap allocated unless if the string is known to be small (which it isn't for this case)

Actually, for many modern std::string series, it won't shrink the actual heap buffer size for change the size smaller, because may some guys want to expand that after some operations, lol. (And the stl implementation want to keep the performance for that scenario. It means resize is basically changing the logical size.)

Kenji Mouri

dongle-the-gadget:
Another thing, wstring has a constructor for C strings, so with the solution I suggested we could easily swap UString with wstring if needed.

MouriNaruto:
> Another thing, wstring has a constructor for C strings, so with the solution I suggested we could easily swap UString with wstring if needed.

Actually, NanaZip additional parts in inherited 7-Zip source code should reduce the dependency of 7-Zip's infrastructure libraries as possible, because we may remove that with rewriting some parts a.k.a. modernization.

Kenji Mouri

MouriNaruto:
But I think maybe can use 7-Zip UString to achieve the goal if @MajThomas knows how to use that. It can satisfied me and resolve @dongle-the-gadget concerns.

Kenji Mouri

MouriNaruto:
@MajThomas

In my opinion, I think the minimum requirement is using std::wstring instead of allocating big block on stack.

You also can choose 7-Zip's UString to achieve that.

Kenji Mouri

PR #723, opened 2025-08-27
Add more targeted editorconfig rules

dongle-the-gadget:
- Enforce Allman style braces
- Suggest PascalCase (although users will need to manually change Visual Studio settings to make it a warning)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -24,6 +24,20 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 vc_generate_documentation_comments = doxygen_slash_star
 
+cpp_new_line_before_open_brace_function = new_line
+cpp_new_line_close_brace_same_line_empty_type = false
+cpp_new_line_close_brace_same_line_empty_function = false
+cpp_new_line_before_else = true
+cpp_new_line_before_catch = true
+cpp_new_line_before_while_in_do_while = true
+
+cpp_naming_rule.general_names.symbols = all_symbols
+cpp_naming_rule.general_names.style = mile_default
+
+cpp_naming_symbols.all_symbols.applicable_kinds = *
+
+cpp_naming_style.mile_default.capitalization = pascal_case
+
 [*.{cs,csx,vb,vbx,asmx}]
 indent_style = space
 indent_size = 4
--- a/NanaZip.sln
+++ b/NanaZip.sln
@@ -55,6 +55,11 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "NanaZip.Codecs.Sfx.Shared",
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "NanaZip.Modern", "NanaZip.Modern\NanaZip.Modern.vcxproj", "{1AB11057-1F48-44D2-A913-C945BFC1AA93}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{8EC462FD-D22E-90A8-E5CE-7E832BA40C5D}"
+	ProjectSection(SolutionItems) = preProject
+		.editorconfig = .editorconfig
+	EndProjectSection
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|ARM64 = Debug|ARM64


NanaZip.sln
```diff
@@ -55,6 +55,11 @@ Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "NanaZip.Codecs.Sfx.Shared",
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "NanaZip.Modern", "NanaZip.Modern\NanaZip.Modern.vcxproj", "{1AB11057-1F48-44D2-A913-C945BFC1AA93}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{8EC462FD-D22E-90A8-E5CE-7E832BA40C5D}"
```
MouriNaruto: I hope we don't include the .editorconfig file in the Visual Studio solution because we may need to contain other files like ReadMe.md if we do that.

.editorconfig
```diff
@@ -24,6 +24,24 @@ insert_final_newline = true
...
+cpp_new_line_before_open_brace_type = new_line
+cpp_new_line_before_open_brace_namespace = new_line
+cpp_new_line_close_brace_same_line_empty_type = false
+cpp_new_line_close_brace_same_line_empty_function = false
+cpp_new_line_before_else = true
+cpp_new_line_before_catch = true
+cpp_new_line_before_while_in_do_while = true
+
+cpp_naming_rule.general_names.symbols = all_symbols
+cpp_naming_rule.general_names.style = mile_default
```
MouriNaruto: If the style name can be customized, I hope it should be "Mile.Project.CodingStyle".

MouriNaruto:
I think you should also create a PR for https://github.com/ProjectMile/Mile.Project.Windows because its .editorconfig will be synchronized to all Nana and Mile series projects.

Kenji Mouri

.editorconfig
```diff
@@ -24,6 +24,24 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 vc_generate_documentation_comments = doxygen_slash_star
```
MouriNaruto: If OK, I hope there should be no empty line.

.editorconfig
```diff
@@ -24,6 +24,24 @@ insert_final_newline = true
...
+cpp_new_line_before_open_brace_block = new_line
+cpp_new_line_before_open_brace_lambda = new_line
+cpp_new_line_before_open_brace_type = new_line
+cpp_new_line_before_open_brace_namespace = new_line
+cpp_new_line_close_brace_same_line_empty_type = false
+cpp_new_line_close_brace_same_line_empty_function = false
+cpp_new_line_before_else = true
+cpp_new_line_before_catch = true
+cpp_new_line_before_while_in_do_while = true
+
```
MouriNaruto: If OK, I hope there should be no empty line.

.editorconfig
dongle-the-gadget: I think it only accepts kebab_case

NanaZip.sln
dongle-the-gadget: It's required for editorconfig to properly work

.editorconfig
MouriNaruto: OK. When I read the document. It seems cpp_naming_rule series are from Visual Studio's Unreal Engine integration extension's feature.
In my opinion, maybe we should only contain Visual Studio's documented .editorconfig definitions: https://learn.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties
It seems mile_default will be used as element reference, so it seems it should be mile_project_cpp will be better.

dongle-the-gadget: It still works even without Unreal Engine installed.

NanaZip.sln
MouriNaruto: Actually, only the IDE configuration UI needs that. The .editorconfig rules seems apply without issues even without include file to solution.

dongle-the-gadget: You're right. I've removed that.

MouriNaruto: I have tested in my Visual Studio 2022 environment, adding this file to the solution view should have no improvement.

.editorconfig
MouriNaruto: OK. It should no issues for that.

MouriNaruto:
Thanks. I have merged that.

I hope you can also create a PR for https://github.com/ProjectMile/Mile.Project.Windows because its .editorconfig will be synchronized to all Nana and Mile series projects.

Kenji Mouri

PR #734, opened 2025-09-01
Add support for CB7 and CBT extensions.

Titokhan:
Similar to CBR and CBZ, these formats, albeit less popular, are renamed 7z and TAR archives for digitized comic books.

### DIFF
--- a/NanaZipPackage/Package.appxmanifest
+++ b/NanaZipPackage/Package.appxmanifest
@@ -101,7 +101,9 @@
               <uap:FileType>.bz2</uap:FileType>
               <uap:FileType>.bzip2</uap:FileType>
               <uap:FileType>.cab</uap:FileType>
+              <uap:FileType>.cb7</uap:FileType>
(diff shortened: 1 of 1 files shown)


dinhngtu:
How common are these formats really?

Titokhan:
CB7 has gained considerable popularity lately, simply because of the higher compression ratio and different compression methods provided by the 7z archive format. Many mainstream comic readers now support CB7 alongside CBZ/CBR, while comic groups often offer releases in the CB7 format instead of CBR.

CBT is not as much popular as compared to the aforementioned three formats, but it's almost trivial to support it at this point. There's also CBA - based on the ACE archive format - but I never come across any such release in the wild.

dinhngtu:
I'm not against CB7. But CBT should probably be removed if it's an uncommon file extension that potentially conflicts with other file types.

Also, please sign off your commits to certify your Developer Certificate of Origin. You can do so by running `git commit --amend --signoff` followed by force pushing.

PR #739, opened 2025-09-03
Implement archive handler restrictions policy

dinhngtu:
Add a policy to control which archive handlers can be loaded by NanaZip.

With this policy, you can significantly limit the attack surface of NanaZip by blocking the parsing of unusual archive formats.

If AllowedHandlers is present, only the handlers specified in said list will be allowed.

If BlockedHandlers is present, the handlers specified in said list will be blocked. BlockedHandlers takes precedence over AllowedHandlers; i.e. handlers specified in BlockedHandlers will still be blocked even if they're specified in AllowedHandlers.

### DIFF
--- a/NanaZip.Core/SevenZip/CPP/7zip/Archive/ArchiveExports.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/Archive/ArchiveExports.cpp
@@ -15,11 +15,105 @@ static unsigned g_NumArcs = 0;
 static unsigned g_DefaultArcIndex = 0;
 static const CArcInfo *g_Arcs[kNumArcsMax];
 
+// **************** NanaZip Modification Start ****************
+#ifndef Z7_SFX
+static PCZZSTR GetMultiStringA(PCSTR Name)
+{
+    PZZSTR Result = nullptr;
+    DWORD Length = 0;
+    LSTATUS Status;
+
+    ::RegGetValueA(
+        HKEY_LOCAL_MACHINE,
+        "Software\\NanaZip\\Policies",
+        Name,
+        RRF_RT_REG_MULTI_SZ,
+        nullptr,
+        nullptr,
+        &Length);
+    if (Length == 0)
+    {
+        return nullptr;
+    }
+
+    Result = static_cast<PZZSTR>(std::calloc(Length, sizeof(CHAR)));
+    if (!Result)
+    {
+        return nullptr;
+    }
+
+    Status = ::RegGetValueA(
+        HKEY_LOCAL_MACHINE,
+        "Software\\NanaZip\\Policies",
+        Name,
+        RRF_RT_REG_MULTI_SZ,
+        nullptr,
+        reinterpret_cast<PVOID>(Result),
+        &Length);
+    if (Status != ERROR_SUCCESS)
+    {
+        std::free(Result);
+        return nullptr;
+    }
+
+    return Result;
+}
+
+static bool MultiStringFindStringA(PCZZSTR Haystack, PCSTR Needle)
+{
+    PCSTR Current;
+
+    if (!Haystack || !Needle)
+    {
+        return false;
+    }
+
+    for (Current = Haystack;
+        *Current;
+        Current = Current + std::strlen(Current) + 1)
+    {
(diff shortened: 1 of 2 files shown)


NanaZip.Core/SevenZip/CPP/7zip/Archive/ArchiveExports.cpp
```diff
@@ -15,11 +15,109 @@ static unsigned g_NumArcs = 0;
...
+    if (!Haystack || !Needle)
+    {
+        return false;
+    }
+
+    for (Current = Haystack;
+        *Current;
+        Current = Current + strlen(Current) + 1)
+    {
+        if (!strcmp(Current, Needle))
```
MouriNaruto: Maybe we should use something like std::strcmp if the std namespace has those standard C functions, or you can use :: to notice that is become the global namespace. (It seems we don't need to make a customized version of those standard C functions in this pull request.)

NanaZip.Core/SevenZip/CPP/7zip/Archive/ArchiveExports.cpp
```diff
@@ -15,11 +15,109 @@ static unsigned g_NumArcs = 0;
...
+            return true;
+        }
+    }
+
+    return false;
+}
+
+static PCZZSTR AllowedHandlers = ([]() -> PCZZSTR
+{
+    return GetMultiStringA("AllowedHandlers");
```
MouriNaruto: I think there are several places lack of `::` to notice that is become the global namespace.

dinhngtu: Does the `::` prefix rule apply to functions in the same TU too?

MouriNaruto: What is the abbreviation for TU?

dinhngtu: Translation unit

MouriNaruto: Got. I think it should be applied to translate unit or wrapper functions because maybe it can reduce some unexcepted implicit behaviors.

NanaZip.Core/SevenZip/CPP/7zip/Archive/ArchiveExports.cpp
dinhngtu: Done, I used `std::` like Mile.Windows.Helpers did.

NanaZip.Core/SevenZip/CPP/7zip/Archive/ArchiveExports.cpp
```diff
@@ -15,11 +15,109 @@ static unsigned g_NumArcs = 0;
...
+        HKEY_LOCAL_MACHINE,
+        "Software\\NanaZip\\Policies",
+        Name,
+        RRF_RT_REG_MULTI_SZ,
+        nullptr,
+        nullptr,
+        &Length);
+    if (Length == 0)
+    {
+        goto fail_query_size;
```
MouriNaruto: I think maybe we should avoid using the goto sentence because it will have some unexpected issues when using C++.

dinhngtu: It should have been safe because there're no initialization order issues. But I converted it to non-goto anyway. Also, I switched to the safer lazy static for `*Handlers` instead of dynamic global init. We should probably do the same for other places (Mitigations etc.)

MouriNaruto: > It should have been safe because there're no initialization order issues.

Actually, I was worried about the future maintenance like if some other guys use C++ stl or other classes do something for that in the future like several years after. It may cause issues. This was why I said that.

MouriNaruto: > We should probably do the same for other places (Mitigations etc.)

I think it seems like a good idea.

PR #758, opened 2025-09-25
Don't reset compression method after changing level

dongle-the-gadget:

### DIFF
--- a/NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
+++ b/NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
@@ -1367,6 +1367,7 @@ bool CCompressDialog::OnCommand(int code, int itemID, LPARAM lParam)
         Get_FormatOptions().ResetForLevelChange();
 
         // **************** NanaZip Modification Start ****************
+        // SetMethod();
         SetMethod(GetMethodID());
         // **************** NanaZip Modification End ****************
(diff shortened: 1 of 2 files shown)


NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
```diff
@@ -1366,7 +1366,9 @@ bool CCompressDialog::OnCommand(int code, int itemID, LPARAM lParam)
       {
         Get_FormatOptions().ResetForLevelChange();
 
-        SetMethod();
+        // **************** NanaZip Modification Start ****************
+        SetMethod(GetMethodID());
```
MouriNaruto, line 1371: Maybe you need to comment out the original code. It's OK if you don't want to modify that because this part will be finally moved into NanaZip.Universal. But I hope you can do that.

dongle-the-gadget: Do you want it before or after the modification mark?

MouriNaruto: Here is the example:

```
// SetMethod();
SetMethod(GetMethodID());
```

dongle-the-gadget: This is done now.

PR #776, opened 2025-11-02
Allow to associate more file types to NanaZip

manfromarce:
This PR adds few file types (that I have tested NanaZip can already open) to the MSIX manifest so that they can be associated to NanaZip in Windows Settings, or NanaZip can be selected from the 'Open with' dialog: 

- .b64: Base64 encoded files, supported by WinZip and WinRAR
- .tz: Tar + ZIP (analogous to .tgz = .tar + .gz)
- .zipx: ZIP with alternative compression algorithms used by WinZip
- .vdi and .vmdk: VirtualBox and VMWare disk images can already be opened by 7-Zip and NanaZip, maybe not all of them but the most common filesystems are supported


dinhngtu (changes requested):
Indeed these are already supported. Though it'd be great if you could add `.cb7` (suggested in #734) at the same time.

Also, please sign off your commits (using `git commit --amend --signoff`) then do a force push.

dinhngtu:
The line endings got messed up on your second commit, for this repo you should enable `core.autocrlf`:

```
git config core.autocrlf true
```

manfromarce:
Ok, it should be fixed now.

Issue #720, opened 2025-08-24
Why is adding files to an existing archive so hard?


uchenzei:
Hello, I understand it might sound like a dumb question but how am I supposed to add files to an existing archive?

If I keep my archive open and then drag new files into it, Nanazip doesn't open the standard compressing options window, so I just can't set any details concerning the new files I'd like to add. Hence this method is useless.

If I keep my archive open and click on the Plus button in the toolbar, it returns the error message "Operation not supported for this folder".

Any suggestion, please?
Thank you

MajThomas:
I ran into that once. Check if you're trying modify a RAR file. Creating or modifying a RAR archive is not possible, you should use WinRAR instead.

uchenzei:
> I ran into that once. Check if you're trying modify a RAR file. Creating or modifying a RAR archive is not possible, you should use WinRAR instead.

Yes I was aware of that, unfortunately this behaviour occurs with any archive type, including standard unencrypted zip. Even if it sounds too strange to me, it really looks like Nanazip doesn't want to let you set your own details for new files being added to an existing archive. 
So how do you people using Nanazip add new files to existing archives? 🤯

dinhngtu:
The Add toolbar button also doesn't work in 7-Zip, so it looks like a bug inherited from there.

dinhngtu:
Okay, I've looked into this. The Add button in 7-Zip simply doesn't support adding files to existing archives (see [7-Zip code](https://github.com/ip7z/7zip/blob/5e96a8279489832924056b1fa82f29d5837c9469/CPP/7zip/UI/FileManager/Panel.cpp#L924-L928)). Although this could be fixed by prompting to add files like how WinRAR does, or to add files by dragging and dropping (which is already supported).

dinhngtu:
Prompt to add files implemented in 6.0.1632.

cachius:
@dinhngtu @MouriNaruto
BEWARE: When adding to an encrypted archive this way without filename encryption, that has been openend to file explorer mode without entering the password, NanaZip Does not ask for the password and silently adds the new files unencrypted! Created https://github.com/M2Team/NanaZip/issues/841 for this

uchenzei:
[...]

PR #809, opened 2026-01-24
Handle Add button when inside archives

dinhngtu:
Make use of CPanel::CopyFromAsk to add files using the Add button.

Unfortunately, there's no easy way to use the add dialog with this as it's part of NanaZipG.
The IFileOpenDialog also had to be manually managed due to the use of multiselect.


cachius:
@MouriNaruto @dinhngtu It's laudable that [this](https://github.com/M2Team/NanaZip/issues/720) was implemented. But [as it stands](https://github.com/M2Team/NanaZip/issues/841) the behavior is worse than before for encrypted archives: It gives the impression that newly added files are encrypted, too, when they are not!

PR #812, opened 2026-01-25
Fix error handling of NanaZip.Codecs

dinhngtu:
- S_FALSE during Open() should also clean up the handler.
- GetArchiveProperty should return VT_EMPTY instead of failing when an archive is not opened, since it will otherwise interfere with update workflows, where UpdateArchive calls GetArchiveProperty on the handler's error props.

### DIFF
--- a/NanaZip.Codecs/NanaZip.Codecs.Archive.DotNetSingleFile.cpp
+++ b/NanaZip.Codecs/NanaZip.Codecs.Archive.DotNetSingleFile.cpp
@@ -475,7 +475,7 @@ namespace NanaZip::Codecs::Archive
 
             } while (false);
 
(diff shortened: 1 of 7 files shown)


NanaZip.Codecs/NanaZip.Codecs.Archive.DotNetSingleFile.cpp
```diff
@@ -475,7 +475,7 @@ namespace NanaZip::Codecs::Archive
 
             } while (false);
 
-            if (FAILED(hr))
+            if (hr != S_OK)
```
MouriNaruto, line 478: Maybe all "hr != S_OK" change to "S_OK != hr" will be better.

Kenji Mouri

PR #823, opened 2026-02-06
misc: 移除更多的未使用文件

peashooter2:
移除更多未使用的文件，以减小二进制文件的大小。


MouriNaruto:
I closed this PR. You should recreate a new PR with the following requirements.

First, we should make the least intrusive modifications (a.k.a. introducing as least modification mark as possible) to the inherited 7-Zip codebase. We can remove unused source files, but referened header files should be kept even it has unused contents.

Also, the pull request title and content should not have any prefix and non-English contents.

Kenji Mouri

PR #825, opened 2026-02-06
Remove more unused files

peashooter2:
Remove more unused files to reduce the size of the binary file.


MouriNaruto:
Note: I merged that without passing the GitHub Actions because I need to verify that locally, which has some changes related to that.

Kenji Mouri

MouriNaruto:
Note: Your PR has passed my local build.

Kenji Mouri

PR #844, opened 2026-02-22
Add support for compressed streaming archives

dinhngtu:
Add 5 new handlers CompressedTar, AutoDeb, AutoDebUncompressed, AutoRpm, AutoRpmUncompressed for decompression of the associated file formats.

All 5 are based on the CompressedStreamArchive handler base and should respect K7 mitigation policies.

CompressedStreamArchive works by caching all inner archive metadata on open, decompressing the inner archive on-the-fly and merging some inner/outer props via GetArchiveProperty. Handlers configure CompressedStreamArchive by passing a CompressedStreamArchiveInfo struct containing outer and inner CArcInfo. Handlers can pick which member in the outer archive to extract (e.g. DEB selects the 2nd member to get the data.tar instead of control.tar)

CompressedStreamArchiveInfo is discovered from within g_Arcs via calling the new function LookupArchiveInfo on CLSIDs provided in CoreExports.hpp. Due to this current dependency on NanaZip.Core internals, the handlers and handler base are all hosted in NanaZip.Core.

Add two new handler flags kCompositeArc and kLongExtension to deal with the new handlers. Particularly, kCompositeArc is needed to bypass these handlers in Open Inside * mode. kLongExtension doesn't seem to really matter but was included for completeness.

Known issues: open progress and open cancelling are not implemented, since CompressedStreamArchive does not monitor the opener progress (to simplify the code; see "﻿Fix CompressedTar management of inner handler" for the removal of the progress monitor)

Tests: https://github.com/M2Team/K7Tests/tree/compressedtar (to be merged after this PR)

﻿


MouriNaruto:
I think I should introduce some changes before I merge this PR, because I want to reduce some dependencies on 7-Zip mainline definitions. For example, we should try to use NanaZip.Specification/NanaZip.Specification.SevenZip.h, we should improve if this lacks some definitions.

Kenji Mouri

dinhngtu:
> I think I should introduce some changes before I merge this PR, because I want to reduce some dependencies on 7-Zip mainline definitions. For example, we should try to use NanaZip.Specification/NanaZip.Specification.SevenZip.h, we should improve if this lacks some definitions.

I finally got around to looking into this. However, this is very difficult because CompressedTar is a special archive handler that looks into the internal CArcInfo registration data. This data is not exposed via IArchive or similar, so I had to use RegisterArc.h internals which itself depends on IArchive.h.

In short, as long as CompressedTar depends on implementation details of NanaZip.Core, I don't think we can transform CompressedTar into a standalone handler using NanaZip.Specification.SevenZip.h just yet.

PR #856, opened 2026-03-01
Use new Astro website

dongle-the-gadget:


Documents/Website/.vscode/extensions.json
dinhngtu, line 1: Let's keep the vscode settings files out of tree.

Documents/Website/src/assets/images/MainWindowDarkMode.png
dinhngtu, line 1: Don't we want to use the same image as the readme? Also, can the image flip with dark mode and/or swap on hover (e.g. using CSS background image attribute)?

Documents/Website/public/FluentSystemIcons-Regular.ttf
dinhngtu, line 1: This is >4MB just for the home/download icons right? Can we get the svg/png and use them instead?

Documents/Website/src/components/StoreBadge/StoreBadge.astro
```diff
@@ -0,0 +1,52 @@
...
+<noscript>
+    <a
+        class="store-badge-link"
+        href={`https://apps.microsoft.com/detail/${productId}?referrer=appbadge&mode=${windowMode}`}
+        target="_blank"
+        rel="noopener noreferrer"
+        data-product-id={productId}
+        data-window-mode={windowMode}
+    >
+        <picture>
```
dinhngtu, line 23: The badge doesn't seem to flip colors in dark mode

Documents/Website/src/assets/images/MainWindowDarkMode.png
dongle-the-gadget: For the first question, I've tried the same image, it made the file listing too small on the screen.

dinhngtu: Agreed, we should have better screenshots and preferably a procedure for capturing those

Documents/Website/src/components/StoreBadge/StoreBadge.astro
dongle-the-gadget: This is on the Microsoft-provided Store badge. I suppose I could write a workaround in JS, but I'll raise this issue to them anyways.

Documents/Website/.vscode/extensions.json
dongle-the-gadget: Those `.vscode` settings were added by Astro templates. IMO since most use VSCode or some type of fork it would be useful to have those in, since those files will recommend the Astro extension and prepare launch profiles for VSCode debugger.

Documents/Website/public/FluentSystemIcons-Regular.ttf
dongle-the-gadget: I don't want to make the component too cluttered with SVG, but loading those fonts are quite slow, yeah. I can replace them.

Documents/Website/src/assets/images/MainWindowDarkMode.png
dinhngtu: dark
light

dongle-the-gadget: The Mica effect didn't show up noticeably on your photos. If that's acceptable then I think it's better that for each mode:
1. We take screenshots in a black background, then a white one
2. Layer mask those two images

This is to get rid of the thin pixel around the image. Might have to write a tool for this.

dinhngtu: Yeah it looked better with 6.0 Preview 2, I'm not sure why. Let me try and retake

dongle-the-gadget: Actually we could remove the background with the Mica on as well, we just have to put a black/white window in the back.

dinhngtu: I had transparency disabled, it looks a bit better when it's reenabled and the screenshot retaken on top of about:blank 

dark
light

dongle-the-gadget: The light mode window still has very noticeable corners, and I'm uncertain which `border-radius` value to use for this. Maybe I'll try the masking approach tomorrow.

dinhngtu: How about using these two themes to highlight the mica effect?

image

dinhngtu: about:blank is not actually black, so here are the properly-screenshot ones:

image

image

image

image

MouriNaruto:
I think I should provide some opinions:

- We should use NanaZip SVG assets from Assets/OriginalAssets folder directly instead of a copy for website source code if OK.
- The light and dark screenshots should be based on the Windows 11 default light and dark bloom wallpaper for future reproduce.
  image
- If you made proper light and dark screenshots, we should replace the repo's markdown document one. Also, you should use png and with `pngquant --force --ext .png --quality 80-80 [The target png]` to optimize size.

Kenji Mouri

dongle-the-gadget:
@MouriNaruto
1. I don't think Astro accepts resources outside of the `src` and `public` folders.
2. Alright
3. In the case of Astro, it probably doesn't matter much since Astro will convert it to WEBP anyways. Sure for Markdown though.

dongle-the-gadget:
It looks like Astro's default image processor is pretty terrible. Using their WebP converter I got horrible color banding


Their AVIF converter is better although still far from original

MouriNaruto:
> @MouriNaruto
> 
> 1. I don't think Astro accepts resources outside of the `src` and `public` folders.
> 2. Alright
> 3. In the case of Astro, it probably doesn't matter much since Astro will convert it to WEBP anyways. Sure for Markdown though.

OK. Only a question left: Can Astro accept the png image? If can, I hope to use png image.

Kenji Mouri

dongle-the-gadget:
Do you mean you want Astro to serve PNG images instead of WebP? It is possible although it will cause us to lose Google search rankings as Google is very tough on WebP use.

Btw, ImageMagick offers lossless compression for WebP which reduces the banding issue a lot. If we want to go WebP then we should make a rule to use Magick or some other lossless compression than use Astro's converter.

MouriNaruto:
> Do you mean you want Astro to serve PNG images instead of WebP? It is possible although it will cause us to lose Google search rankings as Google is very tough on WebP use.

~~Can Astro provide png fallback for output site because some older web browsers may not support webp.~~

It seems OK for modern browsers in past 10 years.

Kenji Mouri

dongle-the-gadget:
> Can Astro provide png fallback for output site because some older web browsers may not support webp.

Sure it's possible. In fact I might prefer AVIF + PNG since Astro's WebP compression is very bad.

MouriNaruto:
image

It seems AVIF needs much newer web browsers. I think your "AVIF + PNG" idea is OK if you want to try.

Also, I hope the light and dark screenshots should be based on the Windows 11 default light and dark bloom wallpaper for future reproduce.
image
Also, you should use `pngquant --force --ext .png --quality 80-80 [The target png]` to optimize size for png images, and copy the new png images to replace the older one in the Documents folder. (I hope markdown images and website images can keep same.)

Kenji Mouri

dongle-the-gadget:
80-80 produces rather bad color banding. I think 85-85 produces a better result.

**80-80**


**85-85**

MouriNaruto:
@dongle-the-gadget 

OK. I think you can leave a note for png compression note in the ReadMe.md of the WebSite source code (or you can leave note in the contribution guide), which will be helpful for us to maintain the website and document in the future.

Kenji Mouri

dongle-the-gadget:
I'm taking the pictures using ShareX. Can I add that in as well? ShareX has transparency settings which removes ugly borders around the window.

MouriNaruto:
> I'm taking the pictures using ShareX. Can I add that in as well? ShareX has transparency settings which removes ugly borders around the window.

I think you can note that if you think you hope to note.

Kenji Mouri

dongle-the-gadget:
@MouriNaruto I've added in the new pictures

Documents/Website/README.md
```diff
@@ -0,0 +1,39 @@
+# NanaZip website source code
```
MouriNaruto: You can simplify the title to `NanaZip Website` because that folder already can make other guys to know it's source code.

Also, the content of this document should keep 80 characters per line for better read experience when people using console or tty.

The file name should be ReadMe.md instead of README.md to keep the same name like other ReadMe documents.

Kenji Mouri

MouriNaruto:
It seems the new NanaZip website is ready. Looks good. Thank you very much.

Kenji Mouri

PR #865, opened 2026-03-06
Add XAML copy dialog

dongle-the-gadget:
Keyboard shortcuts don't currently work yet.


MouriNaruto:
Why you put MileXamlSetXamlContentForContentWindow out of the wrapper function?

Kenji Mouri

MouriNaruto:
Also, we should reduce the dependency on the inherited 7-Zip mainline codebase for the modern parts, to keep the modern parts tidy and simple to maintain.

For example, the browse button for showing the common folder browse dialog should be implemented as the simple Win32 COM calling style in the CopyLocationPage. Also, setting the initial dialog path should be put at the CopyLocationPage constructor.

Kenji Mouri

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,122 @@
...
+            nullptr,
+            IID_IShellItem,
+            InitialFolder.put_void());
+
+        winrt::com_ptr<::IFileDialog> FileDialog =
+            winrt::create_instance<::IFileDialog>(
+                CLSID_FileOpenDialog,
+                CLSCTX_INPROC_SERVER);
+
+        FileDialog->SetOptions(FOS_PICKFOLDERS);
```
MouriNaruto: - We should call GetOptions method first to get the current flags, and then SetOptions.
- We also need to set FOS_FORCEFILESYSTEM if you don't want to treat "This PC", "Network", "Control Panel", or other special shell virtual folder as the result. FOS_DONTADDTORECENT I think can set if we don't want to add the result to the Windows recently access shell folder.

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,122 @@
...
+        FileDialog->SetFolder(InitialFolder.get());
```
MouriNaruto: We should check the SHCreateItemFromParsingName parsing result. We should eliminate the SetFolder call if SHCreateItemFromParsingName returns error.

NanaZip.Modern/NanaZip.Modern.h
```diff
@@ -214,6 +214,58 @@ EXTERN_C INT WINAPI K7ModernShowProgressWindow(
...
+ */
+EXTERN_C LPCWSTR WINAPI K7ModernGetCopyLocationDialogPath(
+    _In_ HWND WindowHandle);
+
+/**
+ * @brief Set the path of the copy location dialog.
+ * @param WindowHandle A handle to the copy location dialog.
+ * @param Path The path to be set.
+ */
+EXTERN_C VOID WINAPI K7ModernSetCopyLocationDialogPath(
```
MouriNaruto: We can remove the unused K7ModernSetCopyLocationDialogPath function.

NanaZip.UI.Modern/SevenZip/CPP/7zip/UI/FileManager/CopyDialog.h
```diff
@@ -12,20 +12,46 @@ const int kCopyDialog_NumInfoLines = 11;
...
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+// **************** NanaZip Modification End ****************
 public:
   UString Title;
   UString Static;
   UString Value;
   UString Info;
   UStringVector Strings;
+// **************** NanaZip Modification Start ****************
+  HWND m_WindowHandle;
```
MouriNaruto: From line 44 to line 46 we should set the initial values to reduce the undefined behaviors.

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,136 @@
...
+        if (SUCCEEDED(::SHCreateItemFromParsingName(
+            this->GetPath(),
+            nullptr,
+            IID_IShellItem,
+            InitialFolder.put_void())))
+        {
+            initialSucceeded = true;
+        }
+
+        winrt::com_ptr<::IFileDialog> FileDialog =
```
MouriNaruto: Only the global namespace functions need to add `::` prefix. Also, for the simple logic which will not throw C++ exceptions (or if has exceptions it must be crashed), we should not use winrt::com_ptr. (Prefer explicit than implicit.)

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,136 @@
...
+            winrt::create_instance<::IFileDialog>(
+                CLSID_FileOpenDialog,
+                CLSCTX_INPROC_SERVER);
+
+        FILEOPENDIALOGOPTIONS Options;
+        FileDialog->GetOptions(&Options);
```
MouriNaruto:
```
DWORD Flags = 0;
if (SUCCEEDED(FileDialog->GetOptions(&Flags)))
{
    FileDialog->SetOptions(
        Flags | FOS_PICKFOLDERS | FOS_FORCEFILESYSTEM)));
}
```

At least, we should have that style.

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,136 @@
...
+        FileDialog->SetOptions(Options | FOS_PICKFOLDERS | FOS_FORCEFILESYSTEM);
+
+        if (initialSucceeded)
+            FileDialog->SetFolder(InitialFolder.get());
```
MouriNaruto, line 116: First we should not eliminate `{` and `}` for `if` and `for` blocks to reduce the future maintenance issues.

Also, I think we should move `SHCreateItemFromParsingName` logic to here to reduce the `initialSucceeded` definitions.

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+        FileDialog->SetTitle(::Mile::WinRT::GetLocalizedString(
+            L"NanaZip.Modern/CopyLocationPage/SetDestinationText",
+            L"Select destination folder.")
+            .c_str());
+
+        if (FAILED(FileDialog->Show(this->m_WindowHandle)))
+            return;
+
+        winrt::com_ptr<::IShellItem> Result;
```
MouriNaruto: Only the global namespace functions need to add `::` prefix. Also, for the simple logic which will not throw C++ exceptions (or if has exceptions it must be crashed), we should not use winrt::com_ptr. (Prefer explicit than implicit.)

NanaZip.Modern/CopyLocationPage.cpp
dongle-the-gadget: Wouldn't this lead to SetOptions not being called if GetOptions doesn't work?

MouriNaruto: I think that I just mentioned an example. The logic style I want to see is more like K7UserModernSHBrowseForFolderW.

MouriNaruto: If we cannot GetOptions properly, maybe that IFileDialog instance has some internal issues. (Because some shell hooks and hacks may add some options.)

MouriNaruto: (Win32 and COM interfaces are really flexible.)

dongle-the-gadget: Could we use default 0 like this:
```cpp
if (FAILED(FileDialog->GetOptions(&Options)))
{
    Options = 0;
}
if (FAILED(FileDialog->SetOptions(Options | FOS_PICKFOLDERS | FOS_FORCEFILESYSTEM))
{
    return; // Not being able to set Folders is a pretty critical error
}
```

NanaZip.Modern/CopyLocationPage.cpp
dongle-the-gadget: `winrt::com_ptr` does not throw exceptions except for `as` and `capture`, neither of which we are using here.
I'm also not a fan of the manual memory management approach of `K7UserModernSHBrowseForFolderW`, since that makes the program fragile.

NanaZip.Modern/CopyLocationPage.cpp
MouriNaruto: > Could we use default 0 like this:
> 
> ```c++
> if (FAILED(FileDialog->GetOptions(&Options)))
> {
>     Options = 0;
> }
> if (FAILED(FileDialog->SetOptions(Options | FOS_PICKFOLDERS | FOS_FORCEFILESYSTEM))
> {
>     return; // Not being able to set Folders is a pretty critical error
> }
> ```

I think this style is OK. And follow the 80 characters per line limit.

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,141 @@
...
+            winrt::create_instance<IFileDialog>(
+                CLSID_FileOpenDialog,
+                CLSCTX_INPROC_SERVER);
+
+        FILEOPENDIALOGOPTIONS Options;
+        if (FAILED(FileDialog->GetOptions(&Options)))
+        {
+            Options = 0;
+        }
+        if (SUCCEEDED(
```
MouriNaruto: I think `SUCCEEDED(` and `FileDialog->SetOptions(` should put at the same line if not exceed the line limit.

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,141 @@
...
+            FileDialog->SetOptions(
+                Options | FOS_PICKFOLDERS | FOS_FORCEFILESYSTEM)))
+        {
+            return;
+        }
+
+        winrt::com_ptr<::IShellItem> InitialFolder;
```
MouriNaruto: Only the global namespace functions need to add `::` prefix.

NanaZip.Modern/CopyLocationPage.cpp
MouriNaruto: That will have a better source code read experience.

NanaZip.Modern/CopyLocationPage.cpp
```diff
@@ -0,0 +1,140 @@
...
+                CLSID_FileOpenDialog,
+                CLSCTX_INPROC_SERVER);
+
+        FILEOPENDIALOGOPTIONS Options;
+        if (FAILED(FileDialog->GetOptions(&Options)))
+        {
+            Options = 0;
+        }
+        if (FAILED(FileDialog->SetOptions(
+                Options | FOS_PICKFOLDERS | FOS_FORCEFILESYSTEM)))
```
MouriNaruto: There are some unnecessary space indents before `Options | FOS_PICKFOLDERS | FOS_FORCEFILESYSTEM)))`.

dongle-the-gadget: Should they line up with FAILED?

MouriNaruto: > Should they line up with FAILED?

I think we should only use 4-space indent like your new commits.

MouriNaruto:
Looks OK, and wait for CI before merging.

Kenji Mouri

PR #869, opened 2026-03-19
Fix mitigation issues in some scenarios. (Suggested by ahmed605.)

dongle-the-gadget:
cc @ahmed605


MouriNaruto:
I will merge that after CI passes. But I will do some modifications after that.

Kenji Mouri

Issue #398, opened 2024-04-08
Introducing NanaZip Preinstall Support


MouriNaruto:
I have posted a tweet for finding OEMs who want to help me request the preinstall permissions for NanaZip. It can make people install NanaZip offline properly because the offline license XML file is needed for installing offline via PowerShell or DISM. (https://twitter.com/MouriNaruto/status/1772202891643936852)

Recently, [Moespirit Tech](https://github.com/Moespirit) reached out and helped me acquire Microsoft Store preinstall permission as the OEM sponsor. I have updated the [Relevant People](https://github.com/M2Team/NanaZip/blob/main/Documents/People.md) document to let NanaZip users know that and show gratitude to them.

Note: The CEO of Moespirit Tech, @SakuraNeko, is one of the NanaZip development team members since the project was created. I helped them create Windows preinstall images for their Intel NUC devices intended for IoT.

I have added the preinstall package for these NanaZip versions:

- [NanaZip 2.0 Update 1 (2.0.450.0)](https://github.com/M2Team/NanaZip/releases/tag/2.0.450)
- [NanaZip 3.0 Preview 0 (3.0.756.0)](https://github.com/M2Team/NanaZip/releases/tag/3.0.756)

For historic versions with Windows 10 Build 17763 support:

- [NanaZip 1.2 (1.2.252.0)](https://github.com/M2Team/NanaZip/releases/tag/1.2)
- [NanaZip 2.0 Preview 1 (2.0.313.0)](https://github.com/M2Team/NanaZip/releases/tag/2.0.313)

Kenji Mouri

Dziubek:
Interesting info about adding the expected functionality:
https://twitter.com/RudyHuyn/status/1781839212935069856

I checked the solution, the installer is downloaded to the disk (nanazip preview installer.exe), I was installed correctly on Windows 11 Pro (I have Microsoft Store)

dongle-the-gadget:
The new EXE installer still requires online access on the machine running it, which doesn’t suit the scenario here.

dongle-the-gadget:
The license file should work across versions, but you may need to update the package.

PR #892, opened 2026-04-23
Update the instructions on how to install NanaZip without an internet connection

gentrydeng:
#398

﻿


MouriNaruto:
I have followed your suggestions, a.k.a. rewriting that part. But I cannot use your PR because it's not suitable for reality.

https://github.com/M2Team/NanaZip/commit/a10ff8ca8599a7a6836d9fe4d77ef6104318149c

So, this PR is closed.

Kenji Mouri

PR #927, opened 2026-06-21
Add "Smart Extract and Open" toolbar button to File Manager

djkcyl:
Adds a one-click File Manager toolbar button that smart-extracts the selected archive (or, when browsing inside an archive, that archive) into the current directory, opens the output folder, and closes the NanaZip window.

Extraction is launched out of process (`showDialog=false`, `smartExtract=true`, `openFolder=true`), so it runs in a separate `NanaZip.Universal.Windows.exe` that opens the output folder on success; closing the File Manager window does not interrupt it.

The button label is localized for en / zh-Hans / zh-Hant; other languages fall back to en.

### DIFF
--- a/NanaZip.Modern/MainWindowToolBarPage.cpp
+++ b/NanaZip.Modern/MainWindowToolBarPage.cpp
@@ -270,6 +270,7 @@ namespace
             Add = 1070,
             Extract = 1071,
             Test = 1072,
+            SmartExtract = 1073, // = kMenuCmdID_Toolbar_SmartExtract (App.h)
             Copy = 546,
             Move = 547,
             Delete = 548,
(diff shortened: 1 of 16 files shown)


MouriNaruto:
- XAML toolbar i18n resources should be modified to `NanaZip.Modern/Strings/*/MainWindowToolBarPage.resw`, not Legacy resources.
- If the button name is `SmartExtractButton`, the enum should be `SmartExtract`, in general, identifier names should be consistent.
- For `‎*/SevenZip/*` changes, you should add the modification marks.

Read https://github.com/M2Team/NanaZip/blob/main/CONTRIBUTING.md before you contribute anything.

Kenji Mouri

djkcyl:
done

MouriNaruto:
```
\NanaZip.Modern\Strings\de\MainWindowToolBarPage.resw
\NanaZip.Modern\Strings\el\MainWindowToolBarPage.resw
\NanaZip.Modern\Strings\en\MainWindowToolBarPage.resw
\NanaZip.Modern\Strings\hu\MainWindowToolBarPage.resw
\NanaZip.Modern\Strings\nl\MainWindowToolBarPage.resw
\NanaZip.Modern\Strings\pl\MainWindowToolBarPage.resw
\NanaZip.Modern\Strings\pt-BR\MainWindowToolBarPage.resw
\NanaZip.Modern\Strings\sq\MainWindowToolBarPage.resw
\NanaZip.Modern\Strings\zh-Hans\MainWindowToolBarPage.resw
```

For other languages you are not good at, you can use English, but we should put the i18n translate item to help other guys update translations.

Also, I think the current NanaZip toolbar width may be too long. I'm afraid that if you add the `Smart Extract and Open` button, some guys will add more, which will cause the toolbar width to get out of control.

Currently, I don't have any better suggestions.

Kenji Mouri

djkcyl:
Thanks for the feedback! I think the toolbar width is still manageable at the moment.

For the long term, if more add-on features (ones that aren't part of upstream 7-Zip) get introduced and the toolbar starts to grow, we could add an option in the Settings menu to let users manage whether these extra toolbar tools are shown. That keeps the default layout clean and the width under control, while still letting users who want the extra buttons opt in.

PR #941, opened 2026-07-06
ElectronAsar: Fix invalid offset used for entry extraction

LukeFZ:
Noticed this while trying to extract an ASAR, seems like this got accidentally changed in 0d98479.

Thanks for creating + maintaining the tool!


MouriNaruto:
You are right. I will merge that after I went home.

Kenji Mouri

PR #952, opened 2026-07-22
New feature: add option to delete archive after extraction

zaum:
Introduce `DeleteArchive` setting (stored in registry) (default: off) to automatically remove archive files upon successful extraction. The deletion occurs only when extraction completes with S_OK. You can turn it on or off in the options.


dinhngtu:
I think this should be an option in the extraction dialog, like the "open folder after extraction" checkbox, rather than a file manager-level one.

zaum:
Maybe it could work there too? 

My goal was to make the process very simple: if "Extract on Open" is enabled and this option is set as well, then double-clicking a ZIP file would simply extract it and delete the file (like on OS X). With as little user intervention as possible.

dinhngtu:
I think there are a few directions to be taken:
- Firstly, the option should rightfully belong in the extraction dialog instead of as an FM setting, as otherwise there's no way to choose the behavior upon trying to extract something.
- Secondly, the option should never permanently delete the original archive, but only move it to Recycle Bin.
- Extracting on open can simply reuse the extraction dialog's previous setting if reducing user interaction is the goal.

NanaZip.Core/SevenZip/CPP/7zip/UI/GUI/ExtractDialog.h
```diff
@@ -1,4 +1,7 @@
-﻿// ExtractDialog.h
+// **************** NanaZip Modification Start ****************
+// (Removed stray BOM character at the start of the original line.)
+// **************** NanaZip Modification End ****************
+// ExtractDialog.h
```
dinhngtu, line 4: This marker should be removed, because it does not impact the actual text. In any case, source files in NanaZip should be using BOM.

NanaZip.Core/SevenZip/CPP/7zip/UI/GUI/ExtractGUI.cpp
```diff
@@ -218,6 +219,8 @@ HRESULT ExtractGUI(
 {
   messageWasDisplayed = false;
 
+  HRESULT result;
+  {
```
dinhngtu, line 223: This needs a marker. Same with the `#include <shellapi.h>`.

NanaZip.UI.Classic/SevenZip/CPP/7zip/UI/FileManager/SettingsPage2.rc
```diff
@@ -18,6 +18,7 @@ BEGIN
   CONTROL  "Want CopyHistory",   IDX_SETTINGS_WANT_COPY_HISTORY,      MY_CHECKBOX, m, 168, xc, 10
   CONTROL  "Want FolderHistory", IDX_SETTINGS_WANT_FOLDER_HISTORY,    MY_CHECKBOX, m, 182, xc, 10
   CONTROL  "Use Lowercase Hashes", IDX_SETTINGS_LOWERCASE_HASHES,     MY_CHECKBOX, m, 196, xc, 10
+
```
dinhngtu, line 21: Looks like this file should be reverted.

NanaZip.Universal/SevenZip/CPP/7zip/UI/FileManager/RegistryUtils.cpp
dinhngtu, line 1: This file needs markers.

dinhngtu (commented):
Sorry for the delay. We're getting there.

@MouriNaruto, what's the current policy for syncing code between the Core/UI/Universal components? 301a147b74cdf72c128ec6fd05339bf73d70a8f4 "Synchronize inherited 7-Zip codebase between NanaZip.Core and NanaZip.Universal" made some unifications between the two code bases involving UI code that did not apply to Core (like the Open folder feature)

MouriNaruto:
@dinhngtu 

> What's the current policy for syncing code between the Core/UI/Universal components?

I think it should be all used source/header/resource file in Core and Universal should be synchronized because it will be easier for future synchronization. (Because we can synchronize Core first, copy Core modified files which also used in Universal to Universal, then synchronize the left parts in Universal, which makes upstream synchronization work much faster.)

UI.Classic and UI.Modern also should synchronize for non XAML UI parts (Although in the future, these will be finally merged in the Universal, lol.)

So, in general:

- Synchronize between Core and Universal.
- Synchronize between UI.Classic and UI.Modern.
- Do not synchronize between Core/Universal and UI.Classic/UI.Modern.

Kenji Mouri

hconps:
Could you consider adding a right-click option called "Delete after Smart Decompression"? This way, when using "Smart Decompression", the source file will be retained, while when using "Delete after Smart Decompression", the source file will be moved to the Recycle Bin. Users can freely choose based on the importance of the file when decompressing it.

Issue #882, opened 2026-04-15
Option to delete archive after (successful) extraction


stderr-to-devnull:
### Describe the feature

Same request as #532 but maybe it won't be completely ignored by the devs this time.

Why is it so hard to add a checkbox for this? And by all means issue a big warning with user confirmation but at least provide the option.

WinRAR is STILL so much better after so many yeas than ALL of the alternatives, for several other features I won't get into here.

### Prerequisite checks

- [x] I have confirmed that this feature isn't yet implemented in the latest Preview release.

zaum:
This is such an ergonomic and essential feature that it's hard to understand why none of the open-source unpackers have implemented it. It's been working this way on OS X for a long time.

Issue #532, opened 2024-12-14
Deleting  ZIP Archive after extraction? How to?


Adish08:
Hey there, I want to delete ZIP/RAR Archive after extraction automatically, how can I do this in NanaZip?

MrHugelberg:
You can't. The only thing that keeps me using WinRAR actually
+1

jsilvermist:
I was using Kubuntu recently where this is supported by default with a "Extract Here and Delete Archive" context menu item, and ever since I can't get over not having this. I'd love it as a feature on NanaZip, even if not enabled in the context menu by default.

Issue #521, opened 2024-12-01
[Request] Extract to folder and delete archive


Miamlya:
A useful time-saving feature would be the ability to automatically export and delete archives. The same operation could also be applied to compression.

PR #964, opened 2026-08-05
Add option to delete the archive after successful extraction

lockewerks:
Implements the request in #882, and also covers #521 and #532.

Extracting an archive never removes it, so download folders fill up with
archives that have already been unpacked. This adds an opt-in option to delete
the source archive once extraction has finished successfully.

## What was added

- A **Delete archive after extracting** checkbox in the extract dialog, next to
  the existing *Open folder after extract* option, persisted under
  `Extraction\DeleteArchiveAfterExtract`.
- A `-sda[-]` command line switch, so the console CLI can use it too.
- `Extract Here` and `Extract to <folder>` never show the dialog, so they read
  the setting the dialog last saved. An explicit `-sda` or `-sda-` on the
  command line wins over the saved setting.

Off by default. All changes are in `NanaZip.Universal` and follow the existing
`OpenFolder` option as a template, so the modification marks stay consistent
with the surrounding inherited code.

## When it deletes

`Extract()` records each archive it opened and decompressed without a fatal
error, including every volume of a multi-volume set, in
`CDecompressStat::ExtractedArcPaths`. The list is only populated when deleting
was requested for a real, non-test extraction.

The GUI and console front ends then delete that list only after their own error
counters confirm the whole batch was clean (`ecs->IsOK()` in the GUI,
`hresultMain == S_OK && !isError && NumFileErrors == 0` in the console). Per
item failures like a CRC error do not fail `archive->Extract()`, so gating on
the front end counters rather than the return code alone is what keeps a
partially damaged archive from being removed. If any archive in a batch reports
an error, nothing in that batch is deleted.

## How it deletes

`SHFileOperationW` with `FOF_ALLOWUNDO` first, so the user can undo it, falling
back to `DeleteFileAlways` for anything the shell did not take. The fallback
matters in two real cases: the shell refuses paths that are not shorter than
`MAX_PATH`, and volumes such as network shares have no Recycle Bin. Without it
the archive would silently stay behind. The operation is silent because the
extraction has already finished and there is no window left to own a prompt.

## Testing

Built `NanaZip.Universal.Console` and `NanaZip.Universal.Windows` for x64
Release. Verified against both binaries:

| Case | Result |
| --- | --- |
| No switch / option off | archive kept |
| `-sda`, saved setting on | archive deleted, files extracted |
| `-sda-` overriding saved setting | archive kept |
| `t` (test) command with `-sda` | archive kept |
| Archive that fails to open | archive kept |
| Archive that opens but has a CRC error | archive kept |
| Multi-volume `.001/.002/.003` | all three volumes deleted |
| Path longer than `MAX_PATH` (287 chars) | deleted through the fallback |
| `Extract Here` and `Extract to <folder>`, no dialog | saved setting honoured |
| Dialog round trip | checkbox saved to the registry and reapplied |

The extract dialog grows by one row (`yc` 168 to 182); the new checkbox sits
below *Open folder after extract* with the buttons unchanged.

Only `Strings/en/Legacy.resw` has the new `Resource3435` string, so the other
locales fall back to English until they are translated.


MouriNaruto:
Before talking the implementation, at least, you should not change the Release Notes documents.

Kenji Mouri

lockewerks:
> Before talking the implementation, at least, you should not change the Release Notes documents.
> 
> Kenji Mouri

Apologies.  It wasn't my intention to touch one of your files.  It was an oversight on my part.  I'm dropping the release notes and force pushing.  Please let me know if there's anything else I need to do so I don't clobber your workflow.

NanaZip.Universal/SevenZip/CPP/7zip/UI/Common/Extract.cpp
```diff
@@ -666,3 +694,48 @@ HRESULT Extract(
...
+       prompt and no useful answer the user could give here. */
+    fo.fFlags = FOF_ALLOWUNDO | FOF_NOCONFIRMATION | FOF_NOERRORUI | FOF_SILENT;
+    ::SHFileOperationW(&fo);
+  }
+
+  /* Whatever the Recycle Bin did not accept, for example a file on a volume
+     that has none, is removed permanently instead of being left behind. */
+  FOR_VECTOR (i, paths)
+    if (NFind::DoesFileExist_Raw(paths[i]))
+      DeleteFileAlways(paths[i]);
```
dinhngtu, line 739: As stated in #952, automatic deletion should never be a destructive action. If moving the archive to Recycle Bin fails, the program should simply skip the deletion instead.

dinhngtu (changes requested):

NanaZip.Universal/SevenZip/CPP/7zip/UI/Common/Extract.cpp
lockewerks: What about an option to enable explicit deletion?  Maybe with an are you sure dialog?  I don't use recycle bin, but I'm just me.  That's not everyone.  If the Recycle bin is enabled, it'll use that.  If, like me, because I'm a maniac, don't like the recycle bin, then it just deletes.  Can split the option to enable recycle bin with no confirmation, or ticking full delete pops up a dialog asking for confirmation and explaining the impact.  Thoughts?

lockewerks: Or perhaps even just an archive destination so instead of removing, the user can specific an archive folder and the original zip goes there instead of getting deleted.

dinhngtu: I think a dialog would be most reasonable and easiest to implement. I don't see the benefit of adding yet another option for such cases.

PR #971, opened 2026-08-19
Add an Extract all button to the copy dialog

dinhngtu:
The behavior of the panel regarding when to show the copy to/extract dialogs is not obvious. So an Extract all button here is a convenient shortcut.

Note, only ‎NanaZip.UI.Modern was changed.

### DIFF
--- a/NanaZip.Modern/CopyLocationPage.cpp
+++ b/NanaZip.Modern/CopyLocationPage.cpp
@@ -8,6 +8,11 @@
 
 #include <string>
 
+namespace winrt
+{
+    using Windows::UI::Xaml::Visibility;
+}
+
 namespace winrt::NanaZip::Modern::implementation
 {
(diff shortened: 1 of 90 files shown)


MouriNaruto:
I think you have used NanaZip.MigrateLegacyStringResources to migrate the string resource. I think you should also commit your modification in `ImmutableArray<ResourceMapping> map` to NanaZip.MigrateLegacyStringResources if I guess it right.

Kenji Mouri

dinhngtu:
> I think you have used NanaZip.MigrateLegacyStringResources to migrate the string resource. I think you should also commit your modification in `ImmutableArray<ResourceMapping> map` to NanaZip.MigrateLegacyStringResources if I guess it right.
> 
> Kenji Mouri

The ExtractAllButton resources were newly-created from scratch instead of being migrated, as I understood that this is the correct way of adding XAML resources.

Issue #321, opened 2023-07-04
[Feature Request] Add password manager with auto-try fill for each zip file with saved passwords


TensorABC:
Add a password manager for nanazip, use label group passwords, try to automatically fill in passwords for compressed packages in specified folders or all areas, and support importing password dictionary blasting

Issue #70, opened 2021-12-10
[Feature Request] Code-page change support


entr0pia:
Supports Code-page change features to fix garbled filename, just like [Bandizip](https://cn.bandisoft.com/bandizip/).

MouriNaruto:
`Try to add language encoding switching support for file names in File Manager.` is the same meaning of `Supports Code-page change features to fix garbled filename`, had been mentioned in the 3.x Series of [Development Roadmap](https://github.com/M2Team/NanaZip#development-roadmap).

So the issue is duplicated.

Kenji Mouri

Artoria2e5:
The roadmap has "Try to add option for save file names with UTF-8 in 7z archives.", which sounds like a mistake. As far as I know the filename encoding in 7z is always UTF-16LE with no room for change. I guess this is instead intended for zip?

PR #973, opened 2026-08-20
Fix UTF-8 passwords for WinZip AES extraction

### LINKED ISSUE #972: ZIP AES-256 archive with a non-ASCII password reports "Wrong password"
### Prerequisite checks

- [x] My system fulfills [the minimum software requirements](https://github.com/M2Team/NanaZip?tab=readme-ov-file#system-requirements).
- [x] I downloaded NanaZip from an official source (Microsoft Store).

### Describe the issue

NanaZip reports `Wrong password` when testing or extracting a WinZip AES-256
ZIP archive whose password contains a non-ASCII character. The password in the
minimal sample contains U+00E4 (`ä`) and is encoded as UTF-8.

The exact same archive and password pass the integrity test with 7-Zip 23.01.
The sample is 218 bytes and contains only one plain-text file named `test.txt`.

### Reproduction steps

1. Open PowerShell.
2. Reconstruct the minimal sample with the following commands:

   ```powershell
   $base64 = 'UEsDBBQAAQBjAJSrFF2zS3U2UgAAADYAAAAIAAsAdGVzdC50eHQBmQcAAQBBRQMAAK+9VVRkGwSeDWj4wWDUF7jn/q0QoAPd6obQUtVcTMuAaZSOgOr1ypzRexlXGHO0IRCS40YsjzdRLAUWHw3Da1ofoOwNTgZBtmwQkT6tt+aJlApQSwECFAMUAAEAYwCUqxRds0t1NlIAAAA2AAAACAALAAAAAAAAAAAA/4EAAAAAdGVzdC50eHQBmQcAAQBBRQMAAFBLBQYAAAAAAQABAEEAAACDAAAAAAA='
   [IO.File]::WriteAllBytes('aes256-unicode-password.zip', [Convert]::FromBase64String($base64))
   Get-FileHash -Algorithm SHA256 'aes256-unicode-password.zip'
   ```

3. Confirm that the SHA-256 value is:

   ```text
   68385160E5AB10B843B777783ED34E0414F35E14B85C72056B5096E39E8128BF
   ```

4. Test the archive with NanaZip using the password `Test-Pässword-2026!`:

   ```powershell
   NanaZipC.exe t '-pTest-Pässword-2026!' '.\aes256-unicode-password.zip'
   ```

5. NanaZip reports:

   ```text
   ERROR: Wrong password : test.txt
   ```

6. Test the same archive with 7-Zip 23.01 and the same password. It reports
   `Everything is Ok`.

### Expected behavior

NanaZip should accept the UTF-8 non-ASCII password and successfully test or
extract `test.txt`, matching 7-Zip's behavior.

### NanaZip veresion

NanaZip 7.0 Preview (7.0.1800.0), x64, installed from Microsoft Store.

### Windows version

Windows 11 Pro, version 25H2, OS build 26200.9168.

### Additional context

The sample was generated with pyzipper 0.3.6 using `WZ_AES`, 256-bit strength,
and `ZIP_STORED`. 7-Zip identifies it as:

```text
Method = AES-256 Store
Characteristics = WzAES : Encrypt
```

This appears distinct from #920, which concerns RAR5 archives rather than a
WinZip AES-encrypted ZIP archive.

wyjh233:
## Summary

- keep the existing ANSI code-page password path as the first choice for
  backward compatibility;
- retry WinZip AES password verification with UTF-8 bytes when the ANSI bytes
  are different and fail verification;
- leave traditional ZipCrypto and PKZIP strong encryption behavior unchanged.

## Root cause

The ZIP handler converted every password from UTF-16 to `CP_ACP` before passing
it to the crypto filter. WinZip AES archives created with UTF-8 password bytes
therefore failed on Windows whenever the non-ASCII password produced different
bytes in the active ANSI code page.

WinZip AES provides a password verification value, so the handler can safely
retain the existing ANSI behavior and retry the same password as UTF-8 only
after the original candidate is rejected.

## Validation

- reproduced the failure with NanaZip 7.0 Preview 7.0.1800.0 and the 218-byte
  sample from #972;
- confirmed that the same archive and password pass with 7-Zip 23.01;
- verified that the source diff passes `git diff --check`;
- completed an x64 Release console build in the
  [fork validation workflow](https://github.com/wyjh233/NanaZip/actions/runs/32377552466);
- tested the built console binary against four WinZip AES-256 archives: the
  UTF-8 issue sample, a UTF-8 password that is also representable in CP936, a
  legacy CP936-encoded password, and an ASCII password; all four passed with
  exit code 0.

Fixes #972

### DIFF
--- a/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
@@ -991,6 +991,10 @@ HRESULT CZipDecoder::Decode(
   bool needCRC = true;
   bool wzAesMode = false;
   bool pkAesMode = false;
+  // **************** NanaZip Modification Start ****************
+  AString_Wipe wzAesUtf8Password;
+  bool wzAesUtf8PasswordIsDefined = false;
+  // **************** NanaZip Modification End ****************
 
   bool badDescriptor = item.IsBadDescriptor();
   if (badDescriptor)
@@ -1114,6 +1118,14 @@ HRESULT CZipDecoder::Decode(
         {
           UnicodeStringToMultiByte2(charPassword, (LPCOLESTR)password, CP_ACP);
         }
+        // **************** NanaZip Modification Start ****************
+        if (wzAesMode)
+        {
+          ConvertUnicodeToUTF8((LPCOLESTR)password, wzAesUtf8Password);
+          wzAesUtf8PasswordIsDefined =
+              (wzAesUtf8Password != charPassword);
+        }
+        // **************** NanaZip Modification End ****************
         /*
         if (wzAesMode || pkAesMode)
         {
@@ -1272,7 +1284,25 @@ HRESULT CZipDecoder::Decode(
         result = _wzAesDecoder->ReadHeader(inStream);
         if (result == S_OK)
         {
+#if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
           if (!_wzAesDecoder->Init_and_CheckPassword())
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+          // **************** NanaZip Modification Start ****************
+          bool passwordIsCorrect =
+              _wzAesDecoder->Init_and_CheckPassword();
+          if (!passwordIsCorrect && wzAesUtf8PasswordIsDefined)
+          {
+            if (_wzAesDecoder->CryptoSetPassword(
+                (const Byte *)(const char *)wzAesUtf8Password,
+                wzAesUtf8Password.Len()) == S_OK)
+            {
+              passwordIsCorrect =
+                  _wzAesDecoder->Init_and_CheckPassword();
+            }
+          }
+          wzAesUtf8Password.Wipe_and_Empty();
+          if (!passwordIsCorrect)
+          // **************** NanaZip Modification End ****************
           {
             res = NExtract::NOperationResult::kWrongPassword;
             return S_OK;
--- a/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
@@ -1315,6 +1315,10 @@ HRESULT CZipDecoder::Decode(
           // **************** NanaZip Modification Start ****************
 #if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
           if (!_wzAesDecoder->Init_and_CheckPassword())
+          {
+            res = NExtract::NOperationResult::kWrongPassword;
+            return S_OK;
+          }
 #endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
           bool IsPasswordCorrect = false;
           if (S_OK == this->_wzAesDecoder->CryptoSetPassword(
@@ -1337,6 +1341,10 @@ HRESULT CZipDecoder::Decode(
           WzAesUtf8Password.Wipe_and_Empty();
           WzAesDefaultEncodingPassword.Wipe_and_Empty();
           if (!IsPasswordCorrect)
+          {
+            res = NExtract::NOperationResult::kWrongPassword;
+            return S_OK;
+          }
           // **************** NanaZip Modification End ****************
           {
             res = NExtract::NOperationResult::kWrongPassword;
--- a/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
@@ -1123,8 +1123,6 @@ HRESULT CZipDecoder::Decode(
 #endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
         if (wzAesMode)
         {
-          // WzAES has no password encoding metadata, so use UTF-8 by default
-          // and preserve CP_ACP as a read-only compatibility fallback.
           ConvertUnicodeToUTF8((LPCOLESTR)password, wzAesUtf8Password);
           UnicodeStringToMultiByte2(
               wzAesCodePagePassword, (LPCOLESTR)password, CP_ACP);
--- a/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
@@ -993,7 +993,7 @@ HRESULT CZipDecoder::Decode(
   bool pkAesMode = false;
   // **************** NanaZip Modification Start ****************
   AString_Wipe WzAesUtf8Password;
-  AString_Wipe WzAesCodePagePassword;
+  AString_Wipe WzAesDefaultEncodingPassword;
   // **************** NanaZip Modification End ****************
 
   bool badDescriptor = item.IsBadDescriptor();
@@ -1127,7 +1127,7 @@ HRESULT CZipDecoder::Decode(
               static_cast<LPCOLESTR>(password),
               WzAesUtf8Password);
           ::UnicodeStringToMultiByte2(
-              WzAesCodePagePassword,
+              WzAesDefaultEncodingPassword,
               static_cast<LPCOLESTR>(password),
               CP_ACP);
         }
@@ -1324,17 +1324,18 @@ HRESULT CZipDecoder::Decode(
             IsPasswordCorrect = this->_wzAesDecoder->Init_and_CheckPassword();
           }
           if (!IsPasswordCorrect &&
-              WzAesCodePagePassword != WzAesUtf8Password)
+              WzAesDefaultEncodingPassword != WzAesUtf8Password)
           {
             if (S_OK == this->_wzAesDecoder->CryptoSetPassword(
-                reinterpret_cast<const Byte *>(WzAesCodePagePassword.Ptr()),
-                WzAesCodePagePassword.Len()))
+                reinterpret_cast<const Byte *>(
+                    WzAesDefaultEncodingPassword.Ptr()),
+                WzAesDefaultEncodingPassword.Len()))
             {
               IsPasswordCorrect = this->_wzAesDecoder->Init_and_CheckPassword();
             }
           }
           WzAesUtf8Password.Wipe_and_Empty();
-          WzAesCodePagePassword.Wipe_and_Empty();
+          WzAesDefaultEncodingPassword.Wipe_and_Empty();
           if (!IsPasswordCorrect)
           // **************** NanaZip Modification End ****************
           {
--- a/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
@@ -993,7 +993,7 @@ HRESULT CZipDecoder::Decode(
   bool pkAesMode = false;
   // **************** NanaZip Modification Start ****************
   AString_Wipe wzAesUtf8Password;
-  bool wzAesUtf8PasswordIsDefined = false;
+  AString_Wipe wzAesCodePagePassword;
   // **************** NanaZip Modification End ****************
 
   bool badDescriptor = item.IsBadDescriptor();
@@ -1115,15 +1115,24 @@ HRESULT CZipDecoder::Decode(
         }
         else
 #endif
+        // **************** NanaZip Modification Start ****************
+#if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
         {
           UnicodeStringToMultiByte2(charPassword, (LPCOLESTR)password, CP_ACP);
         }
-        // **************** NanaZip Modification Start ****************
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
         if (wzAesMode)
         {
+          // WzAES has no password encoding metadata, so use UTF-8 by default
+          // and preserve CP_ACP as a read-only compatibility fallback.
           ConvertUnicodeToUTF8((LPCOLESTR)password, wzAesUtf8Password);
-          wzAesUtf8PasswordIsDefined =
-              (wzAesUtf8Password != charPassword);
+          UnicodeStringToMultiByte2(
+              wzAesCodePagePassword, (LPCOLESTR)password, CP_ACP);
+        }
+        else
+        {
+          UnicodeStringToMultiByte2(
+              charPassword, (LPCOLESTR)password, CP_ACP);
         }
         // **************** NanaZip Modification End ****************
         /*
@@ -1141,6 +1150,8 @@ HRESULT CZipDecoder::Decode(
         }
         */
       }
+      // **************** NanaZip Modification Start ****************
+#if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
       HRESULT result = cryptoSetPassword->CryptoSetPassword(
         (const Byte *)(const char *)charPassword, charPassword.Len());
       if (result != S_OK)
@@ -1148,6 +1159,18 @@ HRESULT CZipDecoder::Decode(
         res = NExtract::NOperationResult::kWrongPassword;
         return S_OK;
       }
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+      if (!wzAesMode)
+      {
+        HRESULT result = cryptoSetPassword->CryptoSetPassword(
+          (const Byte *)(const char *)charPassword, charPassword.Len());
+        if (result != S_OK)
+        {
+          res = NExtract::NOperationResult::kWrongPassword;
+          return S_OK;
+        }
+      }
+      // **************** NanaZip Modification End ****************
     }
     else
     {
@@ -1288,19 +1311,27 @@ HRESULT CZipDecoder::Decode(
 #if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
           if (!_wzAesDecoder->Init_and_CheckPassword())
 #endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
-          bool passwordIsCorrect =
-              _wzAesDecoder->Init_and_CheckPassword();
-          if (!passwordIsCorrect && wzAesUtf8PasswordIsDefined)
+          bool passwordIsCorrect = false;
+          if (_wzAesDecoder->CryptoSetPassword(
+              (const Byte *)(const char *)wzAesUtf8Password,
+              wzAesUtf8Password.Len()) == S_OK)
+          {
+            passwordIsCorrect =
+                _wzAesDecoder->Init_and_CheckPassword();
+          }
+          if (!passwordIsCorrect &&
+              wzAesCodePagePassword != wzAesUtf8Password)
           {
             if (_wzAesDecoder->CryptoSetPassword(
-                (const Byte *)(const char *)wzAesUtf8Password,
-                wzAesUtf8Password.Len()) == S_OK)
+                (const Byte *)(const char *)wzAesCodePagePassword,
+                wzAesCodePagePassword.Len()) == S_OK)
             {
               passwordIsCorrect =
                   _wzAesDecoder->Init_and_CheckPassword();
             }
           }
           wzAesUtf8Password.Wipe_and_Empty();
+          wzAesCodePagePassword.Wipe_and_Empty();
           if (!passwordIsCorrect)
           // **************** NanaZip Modification End ****************
           {
--- a/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
@@ -1284,10 +1284,10 @@ HRESULT CZipDecoder::Decode(
         result = _wzAesDecoder->ReadHeader(inStream);
         if (result == S_OK)
         {
+          // **************** NanaZip Modification Start ****************
 #if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
           if (!_wzAesDecoder->Init_and_CheckPassword())
 #endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
-          // **************** NanaZip Modification Start ****************
           bool passwordIsCorrect =
               _wzAesDecoder->Init_and_CheckPassword();
           if (!passwordIsCorrect && wzAesUtf8PasswordIsDefined)
--- a/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
+++ b/NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
@@ -1123,17 +1123,19 @@ HRESULT CZipDecoder::Decode(
 #endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
         if (wzAesMode)
         {
-          ::ConvertUnicodeToUTF8((LPCOLESTR)password, WzAesUtf8Password);
+          ::ConvertUnicodeToUTF8(
+              static_cast<LPCOLESTR>(password),
+              WzAesUtf8Password);
           ::UnicodeStringToMultiByte2(
               WzAesCodePagePassword,
-              (LPCOLESTR)password,
+              static_cast<LPCOLESTR>(password),
               CP_ACP);
         }
         else
         {
           ::UnicodeStringToMultiByte2(
               charPassword,
-              (LPCOLESTR)password,
+              static_cast<LPCOLESTR>(password),
               CP_ACP);
         }
         // **************** NanaZip Modification End ****************
@@ -1165,9 +1167,9 @@ HRESULT CZipDecoder::Decode(
       if (!wzAesMode)
       {
         HRESULT Result = cryptoSetPassword->CryptoSetPassword(
-            (const Byte *)(const char *)charPassword,
+            reinterpret_cast<const Byte *>(charPassword.Ptr()),
             charPassword.Len());
-        if (Result != S_OK)
+        if (S_OK != Result)
         {
           res = NExtract::NOperationResult::kWrongPassword;
           return S_OK;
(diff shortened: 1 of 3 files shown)


MouriNaruto:
@dinhngtu 

What do you think about this PR?

Kenji Mouri

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1272,7 +1284,25 @@ HRESULT CZipDecoder::Decode(
         result = _wzAesDecoder->ReadHeader(inStream);
         if (result == S_OK)
         {
+#if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
```
MouriNaruto, line 1316: https://github.com/M2Team/NanaZip/blob/main/CONTRIBUTING.md#modifications-for-inherited-7-zip-mainline-source-code

You have not used the multiple-line modifications mark format.

wyjh233: Fixed in 00eafb1. The NanaZip modification start marker now precedes the annotated 7-Zip mainline block, so the replacement follows the documented multiple-line modification format.

dinhngtu:
@wyjh233 @MouriNaruto After some digging, I found that both 7-Zip (26.02 included) and NanaZip currently support "only English letters, numbers and special characters (!, #, $, ...) for password" on ZIP files. Which is to say Unicode support for WZAES-compressed ZIP file passwords is currently simply broken. Why not take the opportunity and convert the entire WZAES path to use only UTF-8, and lift this restriction from compression as well?

dinhngtu:
How the various tools handle passwords containing characters outside of the BMP (e.g. emojis, some CJK characters) would also be an interesting thing to observe.

MouriNaruto:
@dinhngtu I'm agree with your suggestions, but I currently have no idea for achieving that goal because I had not read the implementation too much for WinZip AES implementation inherited from 7-Zip mainline. So, maybe the author of this PR can give some ideas.

Kenji Mouri

wyjh233:
@MouriNaruto @dinhngtu

Thank you both for looking into this.

For context, I reported this issue after receiving an unusual archive that NanaZip could not extract even though the password was correct. Further testing showed that its WzAES password uses UTF-8 bytes, and the patched build can extract it successfully. I cannot share the original archive because it contains third-party content, but the small test fixture reproduces the same behavior.

I agree that NanaZip should use UTF-8 when creating WzAES archives and allow non-ASCII passwords. My only concern is backward compatibility: WinRAR 7.23 on a CP936 Windows system creates WzAES archives using CP936 password bytes, so strict UTF-8-only decoding would make some existing archives unreadable.

Would it be acceptable to always write UTF-8, while reading UTF-8 first and falling back to the current Windows code page? We can also add tests for characters outside the BMP. If the project intentionally prefers strict UTF-8-only decoding, I will follow that direction.

dinhngtu:
> WinRAR 7.23 on a CP936 Windows system creates WzAES archives using CP936 password bytes

Is this the case only on CP936? On a CP1252 system, I tried compressing a file using WinRAR and a complex Unicode password with emojis, and it extracted fine in WinZip. I'm not against fallbacks by themselves, but IMO to avoid changing too much code in the crypto region, we should avoid adding them unless truly necessary.

MouriNaruto:
I'm agree with @dinhngtu, also I want to say something I just recalled that I want to say.

In my opinion, I think it's OK to add UTF-8 and current CP (code page) fallback for extraction, which can improve the compatibility.

But I think maybe we should add an option that lets users can limit passwords to ASCII for ZIP compression to keep interoperability to other archivers. (I'm not against to add UTF-8 password support in compression.)

Kenji Mouri

dinhngtu:
> But I think maybe we should add an option that lets users can limit passwords to ASCII for ZIP compression to keep interoperability to other archivers. (I'm not against to add UTF-8 password support in compression.)

https://github.com/M2Team/NanaZip/blob/38ff0bf1fbbb18089fca9ba3f504e2b7f7d9cda6/NanaZip.Universal/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp#L1244-L1248

For compression, 7-Zip and NanaZip already do that, but I think we should add support for full UTF-8 passwords and turn this error into a warning instead for the WZAES case.

For decompression, a current codepage fallback may be useful in some cases, but its utility is currently somewhat limited, as it's dependent on the user's locale, and there's no easy way to change that in NanaZip. This problem is also relevant for ZIP files whose filenames are encoded in specific locales (cf. #70). So I'd suggest the following sequence:

1. Add support for UTF-8 passwords on the WZAES decompression side as the default;
2. Add support for UTF-8 passwords on the WZAES compression side and make that the unconditional default going forward (without touching PKCrypto);
3. Add CP_ACP as optional support for WZAES decompression (depending on @wyjh233's choice);
  - (Future) Is it worth lifting this to a higher level and let the password decoder try multiple passwords instead? (#321)
4. (Future) Add a mechanism to override CP_ACP usage in NanaZip with a custom codepage.

wyjh233:
Thank you for the guidance. Your CP1252 result prompted me to check this more carefully, and I may not have explained my earlier CP936 test clearly.

On my normal CP936 Windows system, WinRAR 7.23 derived the WzAES key from CP936 password bytes. The password verifier and full authentication code matched CP936, not UTF-8.

I also repeated the WinRAR GUI test under x64 per-process CP1252 locale emulation. Characters representable in CP1252 used CP1252 bytes, while one emoji became `??`; UTF-8 again failed both checks. This may explain the WinZip result if both applications perform the same legacy conversion, although I cannot confirm that for the exact archive you tested.

Given these results, I would prefer to keep step 3 for compatibility with existing WzAES archives. I will update the PR following your proposed scope: UTF-8 by default for WzAES compression and decompression, no PKCrypto changes, and CP_ACP only as a decompression fallback. Please let me know if I have misunderstood anything.

dinhngtu:
> Thank you for the guidance. Your CP1252 result prompted me to check this more carefully, and I may not have explained my earlier CP936 test clearly.
> 
> On my normal CP936 Windows system, WinRAR 7.23 derived the WzAES key from CP936 password bytes. The password verifier and full authentication code matched CP936, not UTF-8.
> 
> I also repeated the WinRAR GUI test under x64 per-process CP1252 locale emulation. Characters representable in CP1252 used CP1252 bytes, while one emoji became `??`; UTF-8 again failed both checks. This may explain the WinZip result if both applications perform the same legacy conversion, although I cannot confirm that for the exact archive you tested.
> 
> Given these results, I would prefer to keep step 3 for compatibility with existing WzAES archives. I will update the PR following your proposed scope: UTF-8 by default for WzAES compression and decompression, no PKCrypto changes, and CP_ACP only as a decompression fallback. Please let me know if I have misunderstood anything.

Let's go with this plan then.

wyjh233:
@dinhngtu @MouriNaruto

Thank you again for the guidance. I have updated the PR following the agreed plan:

- WzAES compression and decompression now use UTF-8 by default.
- Decompression falls back to the current Windows code page for legacy archives.
- PKCrypto behavior remains unchanged.
- The GUI now allows non-ASCII WzAES passwords after a warning and checks the UTF-8 byte limit.

The Windows build passed. I also tested UTF-8 and CP936/CP1252 fallback archives, a non-BMP password, the 99/100-byte boundary, and the existing ZipCrypto behavior.

Would you mind taking another look when convenient? Please let me know if anything still needs adjustment. Thank you.

MouriNaruto:
I have read the current stage implementations. Currently, I have not found the logic issues because maybe I'm not good at that. But I have found some coding style issues. Read https://github.com/M2Team/NanaZip/blob/main/CONTRIBUTING.md#code-style-and-conventions and its mentioned documents for more information.

Kenji Mouri

MouriNaruto:
@dinhngtu What's your opinion on this PR? ~~It seems OK to me.~~ I have found some coding style issues but I mentioned out.

Kenji Mouri

NanaZip.Universal/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
```diff
@@ -1255,6 +1260,34 @@ void CCompressDialog::OnOK()
...
+      {
+        ::ShowErrorMessageHwndRes(*this, IDS_PASSWORD_USE_ASCII);
+        return;
+      }
+      if (::MessageBoxW(
+              *this,
+              ::LangString(IDS_PASSWORD_USE_ASCII),
+              L"NanaZip",
+              MB_OKCANCEL | MB_ICONWARNING) != IDOK)
+        return;
```
MouriNaruto, line 1279: Maybe we should have { and } to wrapper to improve the future maintainability.

NanaZip.Universal/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
```diff
@@ -1255,6 +1260,34 @@ void CCompressDialog::OnOK()
...
+      if (!IsAesMode)
+      {
+        ::ShowErrorMessageHwndRes(*this, IDS_PASSWORD_USE_ASCII);
+        return;
+      }
+      if (::MessageBoxW(
+              *this,
+              ::LangString(IDS_PASSWORD_USE_ASCII),
+              L"NanaZip",
+              MB_OKCANCEL | MB_ICONWARNING) != IDOK)
```
MouriNaruto: - It seems `IDOK != ::MessageBoxW( ...` will be better.
- Wrong indentation for calling parameters.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandlerOut.cpp
```diff
@@ -421,6 +423,25 @@ Z7_COM7F_IMF(CHandler::UpdateItems(ISequentialOutStream *outStream, UInt32 numIt
         if (options.Password.Len() > NCrypto::NWzAes::kPasswordSizeMax)
           return E_INVALIDARG;
       }
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+      if (options.IsAesMode)
+      {
+        if (password)
+          ::ConvertUnicodeToUTF8((LPCOLESTR)password, options.Password);
```
MouriNaruto: Maybe we should have { and } to wrapper to improve the future maintainability.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandlerOut.cpp
```diff
@@ -421,6 +423,25 @@ Z7_COM7F_IMF(CHandler::UpdateItems(ISequentialOutStream *outStream, UInt32 numIt
...
+        if (options.Password.Len() > NCrypto::NWzAes::kPasswordSizeMax)
+          return E_INVALIDARG;
```
MouriNaruto, line 437: Maybe we should have { and } to wrapper to improve the future maintainability.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandlerOut.cpp
```diff
@@ -421,6 +423,25 @@ Z7_COM7F_IMF(CHandler::UpdateItems(ISequentialOutStream *outStream, UInt32 numIt
...
+      }
+      else
+      {
+        if (!IsSimpleAsciiString(password))
+          return E_INVALIDARG;
+        if (password)
+          ::UnicodeStringToMultiByte2(
```
MouriNaruto, line 448: Maybe we should have { and } to wrapper to improve the future maintainability.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandlerOut.cpp
```diff
@@ -421,6 +423,25 @@ Z7_COM7F_IMF(CHandler::UpdateItems(ISequentialOutStream *outStream, UInt32 numIt
...
+      {
+        if (password)
+          ::ConvertUnicodeToUTF8((LPCOLESTR)password, options.Password);
+        if (options.Password.Len() > NCrypto::NWzAes::kPasswordSizeMax)
+          return E_INVALIDARG;
+      }
+      else
+      {
+        if (!IsSimpleAsciiString(password))
+          return E_INVALIDARG;
```
MouriNaruto, line 444: Maybe we should have { and } to wrapper to improve the future maintainability.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1111,9 +1115,28 @@ HRESULT CZipDecoder::Decode(
...
         {
           UnicodeStringToMultiByte2(charPassword, (LPCOLESTR)password, CP_ACP);
         }
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+        if (wzAesMode)
+        {
+          ::ConvertUnicodeToUTF8((LPCOLESTR)password, WzAesUtf8Password);
+          ::UnicodeStringToMultiByte2(
+              WzAesCodePagePassword,
+              (LPCOLESTR)password,
```
MouriNaruto: C++ style casting, please

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1111,9 +1115,28 @@ HRESULT CZipDecoder::Decode(
...
+              CP_ACP);
+        }
+        else
+        {
+          ::UnicodeStringToMultiByte2(
+              charPassword,
+              (LPCOLESTR)password,
```
MouriNaruto: C++ style casting, please

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1129,13 +1152,28 @@ HRESULT CZipDecoder::Decode(
...
       if (result != S_OK)
       {
         res = NExtract::NOperationResult::kWrongPassword;
         return S_OK;
       }
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+      if (!wzAesMode)
+      {
+        HRESULT Result = cryptoSetPassword->CryptoSetPassword(
+            (const Byte *)(const char *)charPassword,
```
MouriNaruto: C++ style casting, please

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1272,7 +1310,33 @@ HRESULT CZipDecoder::Decode(
         result = _wzAesDecoder->ReadHeader(inStream);
         if (result == S_OK)
         {
+          // **************** NanaZip Modification Start ****************
+#if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
           if (!_wzAesDecoder->Init_and_CheckPassword())
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+          bool IsPasswordCorrect = false;
+          if (this->_wzAesDecoder->CryptoSetPassword(
+              (const Byte *)(const char *)WzAesUtf8Password,
```
MouriNaruto: C++ style casting, please

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1272,7 +1310,33 @@ HRESULT CZipDecoder::Decode(
...
+              WzAesUtf8Password.Len()) == S_OK)
+          {
+            IsPasswordCorrect =
+                this->_wzAesDecoder->Init_and_CheckPassword();
+          }
+          if (!IsPasswordCorrect &&
+              WzAesCodePagePassword != WzAesUtf8Password)
+          {
+            if (this->_wzAesDecoder->CryptoSetPassword(
+                (const Byte *)(const char *)WzAesCodePagePassword,
```
MouriNaruto: C++ style casting, please

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1272,7 +1310,33 @@ HRESULT CZipDecoder::Decode(
...
         if (result == S_OK)
         {
+          // **************** NanaZip Modification Start ****************
+#if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
           if (!_wzAesDecoder->Init_and_CheckPassword())
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+          bool IsPasswordCorrect = false;
+          if (this->_wzAesDecoder->CryptoSetPassword(
+              (const Byte *)(const char *)WzAesUtf8Password,
+              WzAesUtf8Password.Len()) == S_OK)
```
MouriNaruto: It seems `S_OK != ...` will be better.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1129,13 +1152,28 @@ HRESULT CZipDecoder::Decode(
...
         res = NExtract::NOperationResult::kWrongPassword;
         return S_OK;
       }
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+      if (!wzAesMode)
+      {
+        HRESULT Result = cryptoSetPassword->CryptoSetPassword(
+            (const Byte *)(const char *)charPassword,
+            charPassword.Len());
+        if (Result != S_OK)
```
MouriNaruto: It seems `S_OK != ...` will be better.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1272,7 +1310,33 @@ HRESULT CZipDecoder::Decode(
...
+          {
+            IsPasswordCorrect =
+                this->_wzAesDecoder->Init_and_CheckPassword();
+          }
+          if (!IsPasswordCorrect &&
+              WzAesCodePagePassword != WzAesUtf8Password)
+          {
+            if (this->_wzAesDecoder->CryptoSetPassword(
+                (const Byte *)(const char *)WzAesCodePagePassword,
+                WzAesCodePagePassword.Len()) == S_OK)
```
MouriNaruto: It seems `S_OK != ...` will be better.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandlerOut.cpp
```diff
@@ -421,6 +423,25 @@ Z7_COM7F_IMF(CHandler::UpdateItems(ISequentialOutStream *outStream, UInt32 numIt
         if (options.Password.Len() > NCrypto::NWzAes::kPasswordSizeMax)
           return E_INVALIDARG;
       }
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+      if (options.IsAesMode)
+      {
+        if (password)
+          ::ConvertUnicodeToUTF8((LPCOLESTR)password, options.Password);
+        if (options.Password.Len() > NCrypto::NWzAes::kPasswordSizeMax)
```
MouriNaruto: Maybe `NCrypto::NWzAes::kPasswordSizeMax < ...` will be better.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandlerOut.cpp
```diff
@@ -421,6 +423,25 @@ Z7_COM7F_IMF(CHandler::UpdateItems(ISequentialOutStream *outStream, UInt32 numIt
...
+          return E_INVALIDARG;
+      }
+      else
+      {
+        if (!IsSimpleAsciiString(password))
+          return E_INVALIDARG;
+        if (password)
+          ::UnicodeStringToMultiByte2(
+              options.Password,
+              (LPCOLESTR)password,
```
MouriNaruto: C++ style casting, please

NanaZip.Universal/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
```diff
@@ -1255,6 +1260,34 @@ void CCompressDialog::OnOK()
...
+        return;
+    }
+    if (IsAesMode)
+    {
+      AString_Wipe Utf8Password;
+      ::ConvertUnicodeToUTF8(this->Info.Password, Utf8Password);
+      if (Utf8Password.Len() > 99)
```
MouriNaruto: What's 99 means? We should have a constant variable instead of magic numbers to improve the readability.
Maybe constant < ... will be better.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1272,7 +1310,33 @@ HRESULT CZipDecoder::Decode(
...
+            {
+              IsPasswordCorrect =
```
MouriNaruto: It seems that it's not necessary to use 2 lines.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1111,9 +1115,28 @@ HRESULT CZipDecoder::Decode(
...
 #endif
+        // **************** NanaZip Modification Start ****************
+#if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
         {
           UnicodeStringToMultiByte2(charPassword, (LPCOLESTR)password, CP_ACP);
         }
+#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
+        if (wzAesMode)
+        {
+          ::ConvertUnicodeToUTF8((LPCOLESTR)password, WzAesUtf8Password);
```
MouriNaruto: C++ style casting, please

MouriNaruto:
Some useful notes:

- Many coding style issues will be found if you use modern Visual Studio (not Visual Studio Code) to write the code, lol.
- Please use a modern IDE instead of an editor.

Kenji Mouri

NanaZip.Universal/SevenZip/CPP/7zip/UI/GUI/CompressDialog.cpp
```diff
@@ -1255,6 +1260,37 @@ void CCompressDialog::OnOK()
...
+          ::LangString(IDS_PASSWORD_USE_ASCII),
+          L"NanaZip",
+          MB_OKCANCEL | MB_ICONWARNING))
+      {
+        return;
+      }
+    }
+    if (IsAesMode)
+    {
+      const UInt32 kWzAesPasswordSizeMax = 99;
```
MouriNaruto: I don't think we should use kWzAes as the local constant define. Also, use Maximum instead of Max.

I think maybe you have not read https://github.com/MouriNaruto/MouriDocs/tree/main/docs/24 mentioned in the contribution guide, at least for this:
> Avoid using abbreviations or acronyms unless they are widely understood.

Which `Max`? [Mad Max](https://en.wikipedia.org/wiki/Mad_Max)?

wyjh233:
Thank you for the detailed review. I am still getting familiar with NanaZip's coding style, and I am sorry for missing several of the project's conventions in the earlier revisions.

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -991,6 +991,10 @@ HRESULT CZipDecoder::Decode(
   bool needCRC = true;
   bool wzAesMode = false;
   bool pkAesMode = false;
+  // **************** NanaZip Modification Start ****************
+  AString_Wipe WzAesUtf8Password;
+  AString_Wipe WzAesCodePagePassword;
```
MouriNaruto: Maybe we can use `WzAesDefaultEncodingPassword` to improve the readability.

Reference:

- https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.default?view=net-10.0#system-text-encoding-default
- https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-multibytetowidechar

MouriNaruto:
It seems that it's OK for me after you have solved all the issues I mentioned.

But you need to wait for the reviews from another NanaZip BDFL that I mentioned. (Read https://github.com/M2Team/NanaZip/blob/main/Documents/People.md#benevolent-dictators-for-life for more information.)

Kenji Mouri

NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1272,7 +1312,32 @@ HRESULT CZipDecoder::Decode(
...
+                    WzAesDefaultEncodingPassword.Ptr()),
+                WzAesDefaultEncodingPassword.Len()))
+            {
+              IsPasswordCorrect = this->_wzAesDecoder->Init_and_CheckPassword();
+            }
+          }
+          WzAesUtf8Password.Wipe_and_Empty();
+          WzAesDefaultEncodingPassword.Wipe_and_Empty();
+          if (!IsPasswordCorrect)
+          // **************** NanaZip Modification End ****************
```
dinhngtu, line 1348: The kWrongPassword branch should be included in its entirety to ease synchronization.

```suggestion
          // **************** NanaZip Modification Start ****************
#if 0 // ******** Annotated 7-Zip Mainline Source Code snippet Start ********
          if (!_wzAesDecoder->Init_and_CheckPassword())
          {
            res = NExtract::NOperationResult::kWrongPassword;
            return S_OK;
          }
#endif // ******** Annotated 7-Zip Mainline Source Code snippet End ********
          bool IsPasswordCorrect = false;
          if (S_OK == this->_wzAesDecoder->CryptoSetPassword(
              reinterpret_cast<const Byte *>(WzAesUtf8Password.Ptr()),
              WzAesUtf8Password.Len()))
          {
            IsPasswordCorrect = this->_wzAesDecoder->Init_and_CheckPassword();
          }
          if (!IsPasswordCorrect &&
              WzAesDefaultEncodingPassword != WzAesUtf8Password)
          {
            if (S_OK == this->_wzAesDecoder->CryptoSetPassword(
                reinterpret_cast<const Byte *>(
                    WzAesDefaultEncodingPassword.Ptr()),
                WzAesDefaultEncodingPassword.Len()))
            {
              IsPasswordCorrect = this->_wzAesDecoder->Init_and_CheckPassword();
            }
          }
          WzAesUtf8Password.Wipe_and_Empty();
          WzAesDefaultEncodingPassword.Wipe_and_Empty();
          if (!IsPasswordCorrect)
          {
            res = NExtract::NOperationResult::kWrongPassword;
            return S_OK;
          }
          // **************** NanaZip Modification End ****************
```

dinhngtu (commented):
Thanks for the great PR. The password changes look good to me. Only a couple small comments remain on markers and synchronization.

- According to https://github.com/M2Team/NanaZip/blob/main/CONTRIBUTING.md#synchronization-between-components, we should sync Core and Universal to ease future synchronization. What's your opinion for this case, @MouriNaruto? Do we want 100% sync between the two or should it be decided case-by-case?
- See further comments below.

MouriNaruto:
@dinhngtu 

image

It seems we don't need 100% sync; we only need to sync if the other side also has the file.

image

For example, if you modify the Core's `SevenZip/CPP/7zip/Archive/IArchive.h`, you also need to change the Universal's `SevenZip/CPP/7zip/Archive/IArchive.h` because the Universal part has the file.

For this PR, we don't need to sync; the first screenshot shows the reason.

Kenji Mouri

Issue #965, opened 2026-08-05
Add "Open target folder after extraction" as a default setting, not just within Extract Files.


kjsc0310:
### Describe the feature

Is it possible to add the default setting of "Open target folder after decompression" in the settings page? Right now this choice only exists in Extract Files, leading to inconvenient operations. Hope this can be optimized.

### Prerequisite checks

- [x] I have confirmed that this feature isn't yet implemented in the latest Preview release.

PR #979, opened 2026-08-26
Open folder after extracting with 'Extract on open'

walaura:
Hi! 👋 

This adds an extra checkbox in the settings page to let the app auto-open the extracted folder after extraction. i really wanted this for myself and I think this also solves #965 partially (only when 'Extract on Open 'is on)

Screenshot 2026-08-26 142902

Verified that the setting works

I'm not super familiar with cpp or old-school windows UI and I'm not super happy with how this looks (ideally this should be indented and go grey when the top option is off) but that's a bit beyond my reach sadly :( this already does what I need so I'm happy for somebody else to run with this PR to polish it


dinhngtu:
We haven't settled the question of whether to use our `OpenFolder` vs `OpnTrgFold` from 7-Zip ZS. cc @MouriNaruto.

PR #988, opened 2026-09-07
Add fix for wrong password on all Zip archives

dinhngtu:


NanaZip.Core/SevenZip/CPP/7zip/Archive/Zip/ZipHandler.cpp
```diff
@@ -1346,10 +1346,6 @@ HRESULT CZipDecoder::Decode(
             return S_OK;
           }
           // **************** NanaZip Modification End ****************
```
MouriNaruto, line 1348: It seems you should care about the modification mark.

dinhngtu: No, in fact the kWrongPassword block is already integrated into the modification block above to avoid tearing the if statement, so this one is a duplicate.

Issue #524, opened 2024-12-01
Always Dark Theme


H5820121:
Could you please add an option for Always Dark Theme, regardless of Windows Theme?
Thank you!

Issue #446, opened 2024-06-06
Option to disable Dark Mode


Dantasstic:
I don't like the look of Dark Mode for NanaZip, so I would like the ability to toggle it off. Can this be added?

modz2014:
put your system in light mode

modz2014:
but yer the background color is too dark and should 255,45,45

Dantasstic:
🙄 I don't want my entire system in light mode, I want this particular app in light mode. Tons of other apps have toggles to follow the system settings or select light or dark. This is pretty standard.

modz2014:
yes i know i will try and push something for it

modz2014:
the project is all over the place now will take me some time to understand it all again

Dantasstic:
> yes i know i will try and push something for it

That would be lovely :)

eggbean:
> yes i know i will try and push something for it

I don't want dark mode in this software either. I can't see any `%APPDATA%` files for NanaZip, so please give information on which registry key the setting uses so I can automatically add it in setup scripts.

VermillionFury007:
I would love this too!!

notNSANE:
adding my overwhelming support for this feature.

Kinda ironic that this feature is useless for me in 95% of apps but when I actually could use it, it doesn't exist 😂

KristoferYG:
yeah,  why the fuk there is no toggle for white mode.... who asked this dark mode?

modz2014:
the annoying thing is you have to build the project without Visual Studio just to test it which is so annoying

modz2014:
I've added a checkbox under the settings for now, but I still need to do more testing. This is because NanaZip's dark mode isn't applied dynamically while the application is running.

modz2014:
@MouriNaruto cant seem to change the darkmode dynamically at runtime also reason why your using dpi

MouriNaruto:
You should see the implementation in https://github.com/M2Team/NanaZip/blob/main/NanaZip.Frieren/NanaZip.Frieren.DarkMode.cpp

Kenji Mouri

modz2014:
@MouriNaruto this is only happening when the system changes from light to darkmode while the application is running not when the button is checked and unchecked

MouriNaruto:
> I've added a checkbox under the settings for now, but I still need to do more testing. This is because NanaZip's dark mode isn't applied dynamically while the application is running.

[...]

PR #996, opened 2026-09-18
Add an in-app inverted dark/light theme for File Manager

HxHedgehog:
This adds an "Invert dark/light mode" option in the file manager
settings, so people stuck on a system theme they don't want for the
app (or just wanting the opposite) can force dark or light without
touching Windows settings. such as https://github.com/M2Team/NanaZip/issues/446 https://github.com/M2Team/NanaZip/issues/524

How it works: - K7User sets the uxtheme preferred app mode and takes care of native
  menus/dialogs, with suspend/resume + an RAII scope for dialogs that
  should keep the native look. - NanaZip.Modern gets K7ModernRefreshTheme and re-applies the theme to
  the XAML islands on switch and on WM _SETTINGCHANGE.
- The state lives in HKCU\Software\NanaZip\FM\InvertTheme.
- File dialogs go through IFileOpenDialog so they follow the theme too.

Nothing under Mile.* was touched. All edits inside SevenZip are wrapped
with the standard NanaZip Modification Start/End markers.
The new string is localized for 14 languages by LLM.

Tested locally on Windows 10 22H2 with a packaged build.
e2e921e64670fd6be5d32bb14009aeca
b1356576e4ee6a4855dfc12598da6a5b
81853221c5301f9dd8eb98a61cad3a6c

﻿


MouriNaruto:
I don't think it's a proper way to implement that goal.

Kenji Mouri

HxHedgehog:
I'm currently satisfied with this version and use it personally. If you have better approach, feel free to edit or refactor it.
Considering this project is a "beautification shell" for 7z kernel, the demand for DL modes switching should be taken seriously—especially since it has existed for a long time and remains unmet.

:-P

